PR #5, opened 2021-02-10
[13.0][ADD] stock_warehouse_partner_security

ChristianSantamaria:
Add a group with reduced permissions for partners in Stock.


stock_picking_security/security/stock_security.xml
```diff
@@ -0,0 +1,82 @@
...
+        <record id="stock.menu_stock_scrap" model="ir.ui.menu">
+            <field name="groups_id" eval="[(4,ref('stock.group_stock_manager'))]"/>
+        </record>
+
+        <record id="stock.menu_procurement_compute" model="ir.ui.menu">
+            <field name="groups_id" eval="[(4,ref('stock.group_stock_manager'))]"/>
+        </record>
+
+        <record id="stock.group_stock_manager" model="res.groups">
+            <field name="implied_ids" eval="[(6, 0, [ref('stock_picking_security.group_stock_picking_partner'), ref('account.group_account_invoice')])]"/>
```
dalonsod: Tried to install, faced an error:
```
"External ID not found in the system: account.group_account_invoice" while evaluating
"[(6, 0, [ref('stock_picking_security.group_stock_picking_partner'), ref('account.group_account_invoice')])]
```
And that's right, making this addon depending on `stock`, `account` addon shouldn't be used.

You'd get the desired behavior changing this (maybe you should test it and check if `account` permissions are kept):

```suggestion
            <field name="implied_ids" eval="[(4,ref('stock_picking_security.group_stock_picking_partner'))]"/>
```

dalonsod (changes requested):
See comment. And commit history is not in a proper way yet, but it could be fixed later, don't worry

dalonsod:
@lmiguens-solvos what do you think about addon name (`stock_picking_security`)?

lmiguens-solvos:
> @lmiguens-solvos what do you think about addon name (`stock_picking_security`)?

Well, it is not very explicit about what it provides at the security level... does not indicate that I will only have permission for operations in my warehouses
But it may be too long to put something like: stock_picking_my_warehouses_restriction_security ??

dalonsod:
Well, I'd focus to `picking` term. The current changes are the following:
- The `Inventory` group hierarchy is changed: there's a new group inserted between `Users` a `Administrator` groups.
- `User` group now has lost several menu accesses, those ones are moved to `Administrator` group.

These changes have an scope greater than picking operations. Thus, we're redefining stock security schema a little bit...

lmiguens-solvos:
> Well, I'd focus to `picking` term. The current changes are the following:
> 
> * The `Inventory` group hierarchy is changed: there's a new group inserted between `Users` a `Administrator` groups.
> * `User` group now has lost several menu accesses, those ones are moved to `Administrator` group.
> 
> These changes have an scope greater than picking operations. Thus, we're redefining stock security schema a little bit...

ok, perhaps a name like stock_inventory_security would be better??

lmiguens-solvos (changes requested):
- It is also necessary to add access to the menu "Reports" -> "Inventory report" to make inventory adjustments from there
- Also add that this user can see the operations of the contact parent_id (user contact parent_id warehouse operations)

dalonsod:
> ok, perhaps a name like stock_inventory_security would be better??

`inventory` and `stock` are synonyms, at least in Odoo... maybe simply `stock_security`?

lmiguens-solvos:
> > ok, perhaps a name like stock_inventory_security would be better??
> 
> `inventory` and `stock` are synonyms, at least in Odoo... maybe simply `stock_security`?



> > ok, perhaps a name like stock_inventory_security would be better??
> 
> `inventory` and `stock` are synonyms, at least in Odoo... maybe simply `stock_security`?

And something more extensive that indicates the limitation introduced with this addon? 
something like stock_warehouse_partner_limited_security

dalonsod:
> And something more extensive that indicates the limitation introduced with this addon?
> something like stock_warehouse_partner_limited_security

OK, I propose `stock_warehouse_partner_security`, without "limited" :-)

Maybe this addon should be improved in the future with more groups with other requisites, but not right now.

dalonsod:
Closed and re-opened in #8

PR #6, opened 2021-02-12
[13.0] [FIX] stock_picking_auto_create_lot_product_supplier_date_eco: create lot if not exists to avoid possible duplicate restrictions

lmiguens-solvos:
Create lot if not exists to avoid possible duplicate restrictions


dalonsod:
Tested, LGTM. Done the following changes:
- Bump minor version on manifest according to a fix.
- Renamed commit from improvement to fix.
- cosmetic change on assignment (added spaces)

Merge is coming

PR #7, opened 2021-02-18
[13.0] [IMP] stock_picking_auto_create_lot_product_supplier_date_eco

lmiguens-solvos:
Add new process to create intermediate transfers (generating stock in the new lot that is automatically generated from the stock in the initial lot (declared))


dalonsod (approved):
Merging

PR #8, opened 2021-02-23
[13.0][ADD] stock_warehouse_partner_security

dalonsod:
Includes #5 and these new improvements and fixes:
- Pickings: instead of showing pickings depending on source and dest locations, pickings shown are those that belongs to allowed picking types for `Partners` group.
- Inventory report: prevent inventory quants creation from locations not allowed making permission checks within `create()` method (creation rules are bypassed in `stock` addon). Every location is listed to be selected, anyway.
- Rebase made in order to prevent merge conflicts with `13.0` branch.


dalonsod:
>     * Pickings: instead of showing pickings depending on source and dest locations, pickings shown are those that belongs to allowed picking types for `Partners` group.

@lmiguens-solvos do you agree, or some problems are expected with this change?

dalonsod:
@ChristianSantamaria could you test? Thanks!

lmiguens-solvos:
> > ```
> > * Pickings: instead of showing pickings depending on source and dest locations, pickings shown are those that belongs to allowed picking types for `Partners` group.
> > ```
> 
> @lmiguens-solvos do you agree, or some problems are expected with this change?

I think it may not achieve what you want. As the purchase orders will be made with the type of reception operation of the main warehouse, originating from the user's warehouse.

dalonsod:
OK, I'll revert the change

dalonsod:
New changes made:
* Partners now can only access to Transfers and Inventory Report. Dashboard access is removed.
* A new menu is created (_My operations_), with the two accesses described above. This menu is also avaliable for Inventory Administrator users
* _My Contact Transfers_ only shows those pickings whose contact is the related contact of the current user. Creating pinckings is disabled from this menu.
* _Inventory Report_ keeps the same functionality described in this PR, without major changes.

@lmiguens-solvos could you check?

lmiguens-solvos:
> * Partners now can only access to Transfers and Inventory Report. Dashboard access is removed.

They can access to Contacts menu, in which they have access to create a new contact and modify existing ones. I think it shouldn't. The other menus (Discuss, Calendar and Website) I think they have no problem because they are
In the Inventory menu they can access too to Master Data... is it possible to remove access from that menu?
Although they can then access an product from a picking, but they can't modify it, and that's fine.. 
Within the product edition, could access to the rules and routes buttons be removed?

> * A new menu is created (_My operations_), with the two accesses described above. This menu is also avaliable for Inventory Administrator users

Good idea

> * _My Contact Transfers_ only shows those pickings whose contact is the related contact of the current user. Creating pinckings is disabled from this menu.

checked

> * _Inventory Report_ keeps the same functionality described in this PR, without major changes.

checked

dalonsod:
> They can access to Contacts menu, in which they have access to create a new contact and modify existing ones. I think it shouldn't. The other menus (Discuss, Calendar and Website) I think they have no problem because they are

Those menus belongs to other applications, and those permissions should be discussed in other addons. Contacts write menu access is actually this group configuration:

imagen

> In the Inventory menu they can access too to Master Data... is it possible to remove access from that menu?

Not seen for me:

imagen

Try unistall and, then reinstall addon and reassign user to `Partner` group.

> Although they can then access an product from a picking, but they can't modify it, and that's fine..
Within the product edition, could access to the rules and routes buttons be removed?

OK, I'll check it

lmiguens-solvos:
> Those menus belongs to other applications, and those permissions should be discussed in other addons. Contacts write menu access is actually this group configuration:

ok, you are right. Sorry!



> Not seen for me:

I did what you told me: uninstalled and reinstalled, but it's the same, 
maybe I have more addons installed, but I can't find where to remove these permissions

image

dalonsod:
> Within the product edition, could access to the rules and routes buttons be removed?

I'm experiencing some problems trying this. These buttons are added by the following view (id. `product_form_view_procurement_button`):

https://github.com/odoo/odoo/blob/13.0/addons/stock/views/product_views.xml#L195

- _Reordering rules_ button (L257) could be hidden adding `group_stock_manager` to `groups`.
- but, both _Routes_ (L267) and _Putaway Rules_ (L273) buttons have already a `groups` configuration (`group_stock_multi_locations`), so adding the manager group won't take effect. Using `invisible` attribute doesn't appear to be possible here.

Looking at the whole view (enabled form `group_stock_user`), is it possible to disable it for Manager group, or we actually need some content of it?

lmiguens-solvos:
> Looking at the whole view (enabled form `group_stock_user`), is it possible to disable it for Manager group, or we actually need some content of it?

There would be no problem if no one had access to these buttons.
We would have access to the routes through the corresponding inventory menus.

dalonsod:
> There would be no problem if no one had access to these buttons.
We would have access to the routes through the corresponding inventory menus.

OK, view moved from `Users` to `Administrator` role.

Also added optional flag to hide `partner_id` column on tree view when the menu is the "My Contact Transfers" one.

lmiguens-solvos:
> > There would be no problem if no one had access to these buttons.
> > We would have access to the routes through the corresponding inventory menus.
> 
> OK, view moved from `Users` to `Administrator` role.
> 
> Also added optional flag to hide `partner_id` column on tree view when the menu is the "My Contact Transfers" one.

ok, I saw it and I think it can be like this

dalonsod:
> I did what you told me: uninstalled and reinstalled, but it's the same,
maybe I have more addons installed, but I can't find where to remove these permissions

At last fixed, thanks for the help

lmiguens-solvos (approved):
LGTM

dalonsod:
Squashed commits, proceeding to merge...

PR #10, opened 2021-03-04
[13.0][ADD] stock_picking_eco_tag

ChristianSantamaria:
Adds Eco Pdf Report in Stock Picking.

Translations missing


stock_picking_eco_tag/report/stock_picking_eco_tag_template.xml
```diff
@@ -0,0 +1,38 @@
...
+                        <t t-foreach="move.move_line_ids" t-as="move_line">
+                          <div style="display: inline-table; height: 160px; width: 367px; border: 1px solid black;">
+                            <div class="d-flex">
+                                <div style="display: -webkit-box; display: -webkit-flex; display: flex; -webkit-box-align: center; justify-content: center; align-items: center; margin-left: 3px; width: 35%">
+                                    <img t-attf-src="data:image/jpg;base64,{{ move.company_id.logo }}" style="width: 130px"/>
+                                </div>
+                                <div style="width: 65%; padding-left: 6px;" t-if="picking.move_lines">
+                                    <div style="padding-top: 5px;">
+                                        <p class="mb-0" style="font-size: 10px">Product: <t t-esc="move_line.product_id.name"/></p>
+                                        <p class="mb-0" style="font-size: 10px">Var / Cat: <t t-esc="move_line.product_id.pos_categ_id.name"/></p>
```
dalonsod: Wrong reference to a field (`pos_categ_id`) not linked in addon dependencies, if `point_of_sale` is not installed an error is fired. Please try `categ_id` instead.

dalonsod (changes requested):

stock_picking_eco_tag/report/stock_picking_eco_tag_template.xml
```diff
@@ -0,0 +1,38 @@
...
+                          <div style="display: inline-table; height: 160px; width: 367px; border: 1px solid black;">
+                            <div class="d-flex">
+                                <div style="display: -webkit-box; display: -webkit-flex; display: flex; -webkit-box-align: center; justify-content: center; align-items: center; margin-left: 3px; width: 35%">
+                                    <img t-attf-src="data:image/jpg;base64,{{ move.company_id.logo }}" style="width: 130px"/>
+                                </div>
+                                <div style="width: 65%; padding-left: 6px;" t-if="picking.move_lines">
+                                    <div style="padding-top: 5px;">
+                                        <p class="mb-0" style="font-size: 10px">Product: <t t-esc="move_line.product_id.name"/></p>
+                                        <p class="mb-0" style="font-size: 10px">Var / Cat: <t t-esc="move_line.product_id.categ_id.name[3:]"/></p>
+                                        <p class="mb-0" style="font-size: 10px">Net weight: <t t-esc="move_line.qty_done"/> <t t-esc="move_line.product_id.uom_id.name"/></p>
```
dalonsod: Missing numer format. A custom format related to user or picking partner should be required. @lmiguens-solvos 
- how many decimals should we need?
- lang format, related to Odoo user, or picking partner?

lmiguens-solvos: - We need 3 decimals. But this precision maybe from self.env['decimal.precision'].precision_get('Product Unit of Measure')
- About the language, I'm not sure, but maybe it should be the company language or picking partner... but it is possible tags should always go in one language?

dalonsod: > * We need 3 decimals. But this precision maybe from self.env['decimal.precision'].precision_get('Product Unit of Measure')

I think it should be possible. @ChristianSantamaria could you try?

> * About the language, I'm not sure, but maybe it should be the company language or picking partner... but it is possible tags should always go in one language?

It's possible creating a `res.config.settings` parameter, with an extra stuff. If you prefer it rather than the other options, that shoud be the way IMO.

lmiguens-solvos: On second thought ... perhaps the best may be the language of the picking.partner, right? 
So whoever receives the label, understand it (assuming it is generated in the picking of the sales order)

ChristianSantamaria: @dalonsod Yes of course, i start now

dalonsod: Then,
- quantity number format according picking partner language.
- decimal precision according @lmiguens-solvos requirements

stock_picking_eco_tag/report/stock_picking_eco_tag_template.xml
```diff
@@ -15,7 +15,8 @@
                                     <div style="padding-top: 5px;">
                                         <p class="mb-0" style="font-size: 10px">Product: <t t-esc="move_line.product_id.name"/></p>
                                         <p class="mb-0" style="font-size: 10px">Var / Cat: <t t-esc="move_line.product_id.categ_id.name[3:]"/></p>
-                                        <p class="mb-0" style="font-size: 10px">Net weight: <t t-esc="move_line.qty_done"/> <t t-esc="move_line.product_id.uom_id.name"/></p>
+                                        <t t-set="decimal_precision" t-value="request.env['decimal.precision'].precision_get('Product Unit of Measure')"/>
```
dalonsod: Is it possible to place this line outside `t-foreach` loop?

stock_picking_eco_tag/report/stock_picking_eco_tag_template.xml
```diff
@@ -15,7 +15,8 @@
...
+                                        <p class="mb-0" style="font-size: 10px">Net weight: <t t-esc="move_line.qty_done" t-options='{"widget": "float", "precision": decimal_precision}'/> <t t-esc="move_line.product_id.uom_id.name"/></p>
```
dalonsod: language selection for the unit of measure is not working. Changed partner language from `es` to `en`, this value remains in `es` language (the Odoo user language, probably). Maybe `t-esc` is the problem here.
Please try this (not tested):
```suggestion
                                        <p class="mb-0" style="font-size: 10px">Net weight: <t t-esc="move_line.qty_done" t-options='{"widget": "float", "precision": decimal_precision}'/> <span t-field="move_line.product_id.uom_id.name"/></p>
```

dalonsod (changes requested):
See comments

dalonsod:
@ChristianSantamaria if possible, please attach an screenshot with an example of the PDF result (with more than one line of tags, if possible), then @lmiguens-solvos could inspect it without locally testing the code

lmiguens-solvos:
> @ChristianSantamaria if possible, please attach an screenshot with an example of the PDF result (with more than one line of tags, if possible), then @lmiguens-solvos could inspect it without locally testing the code

LGTM. Except maybe the PDF margins ... But since the exact size is still pending ... right?
image

ChristianSantamaria:
@lmiguens-solvos have been faster than me :), and in the case of the page change it would be such that:
imagen

stock_picking_eco_tag/report/stock_picking_eco_tag_template.xml
ChristianSantamaria: Done

stock_picking_eco_tag/report/stock_picking_eco_tag_template.xml
dalonsod: This is still not working, only if user language is changed, that description is changed. @lmiguens-solvos maybe this issue needs some further investigation, and could be fixed later anyway. What do you think, is it ready to be merged?

lmiguens-solvos: Yes, I think you can merge it even if that detail remains

dalonsod (approved):
LGTM
@ChristianSantamaria  please squash commits

dalonsod:
Merging

PR #14, opened 2021-03-10
[13.0][ADD] stock_picking_move_line_descriptions

ChristianSantamaria:
Adds Description column of Purchase in Picking
imagen


stock_picking_move_line_descriptions/models/stock_move_line.py
```diff
@@ -0,0 +1,10 @@
+# © 2021 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (http://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import fields, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+    
+    description = fields.Char('Description', index=True, related='move_id.name', required=True)
```
dalonsod: Suggested:
```suggestion
    description = fields.Char(related='move_id.name')
```
- `string` attribute is automatically resolved by Odoo
- `index` is not required here IMO
- since this field is related read only, required has no sense.

stock_picking_move_line_descriptions/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Adds Description column of Purchase in Picking
```
dalonsod: Suggested (description is for move lines; purchase order lines is only one of the possible origins):
```suggestion
Adds Description column to Picking Detailed Operations
```
The same in `__manifest__` file

dalonsod (commented):
Only code review

dalonsod (approved):
LGTM
@lmiguens-solvos maybe you could test in your local environment and verify if fills your requirements in a real situation

lmiguens-solvos:
> LGTM
> @lmiguens-solvos maybe you could test in your local environment and verify if fills your requirements in a real situation

LGTM.
This description is in purchase an sale picking, not only in purchase. 
It doesn't matter, it can be like this. Although it was necessary in purchases, and in sales it is surely repetitive with the name of the product, but it does not matter.
I tested in my environment, and after modifying my process of automatic generation of detailed operations (in which at first I did not put the move_id) and then going to put the move_id in order to have the description, it seems that everything is still fine and show the description that they need. 
Then, for me is ok, too.

dalonsod:
> This description is in purchase an sale picking, not only in purchase.
It doesn't matter, it can be like this. Although it was necessary in purchases, and in sales it is surely repetitive with the name of the product, but it does not matter.

Yes, since are several types and origins for pickings, maybe could be parametrized. Maybe in other improvement of this addon, if needed.

@ChristianSantamaria please squash commits and I'll merge it

PR #16, opened 2021-03-30
[13.0] [IMP] stock_picking_eco_tag: Add Craega logo

ChristianSantamaria:
Tag style update by adding the Craega logo.

When adding the new logo I found that the product with the name `ESPINAFRE PERPETUA / REMOLACHA DE FOLLA` broke the structure of the tag, everything had to be repositioned, as well as the font size to 9px.
There are two possible styles for tags: 

1. This is the current updated tag, the product with the long name becomes in two rows but two spaces are generated above and below the Biochousa logo.
imagen

2. With this option the product could be in only one line (for a while, until they create a product with a longer name), on the other hand, if the name is short, an empty space is created too large. And by placing the ecological logo under biochousa, it creates a space on its right. 
imagen

New style proposals are welcome


dalonsod:
OK, I'll test it locally ASAP. Regarding design questions, if (1) is correct for @lmiguens-solvos , I'd preserve it. Since we don't know if size is the final one, the design probably should be changed one more time, so we won't spend more time on it

stock_picking_eco_tag/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.0.1",
```
dalonsod: This is an improvement, not a fix, so I'd prefer this version bump:
```suggestion
    "version": "13.0.2.0.0",
```

dalonsod (changes requested):
See comment

lmiguens-solvos:
> OK, I'll test it locally ASAP. Regarding design questions, if (1) is correct for @lmiguens-solvos , I'd preserve it. Since we don't know if size is the final one, the design probably should be changed one more time, so we won't spend more time on it

Yes, for me (1) is correct. I do not dislike the product with the long name becomes in two rows.
Perhaps logo CRAEGA too small, but I know that the space is very tight... 
An idea, I think the same width for Craega log as the leaf logo would be better, and move the address and other data of biochousa under the Biochousa logo... and the Biochousa logo adjusted at the top left... 
Well...  one label, multiple design ideas... 
But as David says: Since we don't know if size is the final one, the design probably should be changed one more time, so we won't spend more time on it...

ChristianSantamaria:
Due to the customer change the tags had to be redesigned to a vertical ticket style with a width of 8cm. The value of the attribute "CATEGORÍA" was also added in category.
Now the style of the tags are very simple, it is like this (8cm width x 10.5cm height):
imagen
But I also had other designs not so ticket like this: 
imagen
What do you think?

lmiguens-solvos:
> Due to the customer change the tags had to be redesigned to a vertical ticket style with a width of 8cm. The value of the attribute "CATEGORÍA" was also added in category.
> Now the style of the tags are very simple, it is like this (8cm width x 10.5cm height):
> imagen
> But I also had other designs not so ticket like this:
> imagen
> What do you think?

Well, a difficult decision,.. 
in a first visual I like the first one better, 
but it is clear that in the second the space is better used

dalonsod:
@ChristianSantamaria IMO second design is better. Maybe it need a minor adjustment ("From" line has no height enough)

stock_picking_eco_tag/models/stock_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def lines_category_product(self):
+
+        for attribute in self.product_id.attribute_line_ids:
+            #if attribute.id = self.env.ref("stock_picking_eco_tag.product_attribute_category").id
+            if attribute.attribute_id.name == "CATEGORÍA":
+                if attribute.value_ids.name:
+                    return attribute.value_ids.name
+
+        return False
```
dalonsod: Is it possible to optimize this code like hre: https://github.com/solvosci/slv-stock/blob/13.0/stock_picking_deliveryslip_subtotals/models/stock_picking.py#L10 ?

If we have a picking with a lot of lines with the same product, the code becomes inefficient

And please protect code, because it will fail if that attribute has more than a value

dalonsod (changes requested):
See comment

ChristianSantamaria:
New updated tag design (8cm x 7.5cm)
Here are two examples of tags with different name sizes:
imagen
imagen

dalonsod (approved):
Added code protection, now category attribute is created if not exists yet when installing.

Tested, LGTM. @ChristianSantamaria please squash commits

PR #17, opened 2021-04-12
[13.0] [FIX] stock_picking_eco_tag: page adjustment and remove last blank page

ChristianSantamaria:
Fixed the bug of the extra blank page and the size of the tag

### DIFF
--- a/stock_picking_eco_tag/__manifest__.py
+++ b/stock_picking_eco_tag/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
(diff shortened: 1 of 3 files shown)


dalonsod (changes requested):
Please reword commit message with something like "[FIX] stock_picking_eco_tag: page adjustment and remove last blank page".

Why did you solve extra blank page issue?

ChristianSantamaria:
Apparently the vertical size of the page format did not adjust well to the height of the tag, a small space had to be left so that the last blank page would not be generated.
imagen

dalonsod (approved):
OK, I'll merge it :)

PR #19, opened 2021-07-12
[13.0][ADD] stock_picking_mgmt_weight, stock_picking_mgmt_weight_pot

ChristianSantamaria:
Main features:
- Add the vehicles field to the stock.picking and res.partner model
- Add the fields to the stock.move model:
   - Theoretical quantity demanded
   - Tare
   - Gross
   - Net
   - "Tare and Net Date"
- Add new filters in res.partner:
   - With Vehicles
   - Without Vehicles
   - Search by license plate
- Add filters in stock.picking:
  - Search by license plate
  - Add the vehicle field to the tree view

Pending changes:
- Check the domain in stock.picking vehicles.
- Give values ​​in a function to the Tare and Net date fields
- Create an operation in data with the new characteristic if it is a scale.


stock_picking_mgmt_weight/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# © 2021 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+{
+    "name": "Stock Picking Move Weight",
+    "summary": """
+
```
dalonsod: Pending summary

stock_picking_mgmt_weight/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import models, fields
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    theoretical_qty = fields.Float(string='Theoretical quantity demanded')
+    tare = fields.Float(string='Tare')
+    date_tare_weight = fields.Date(string='Tare weight date')
```
dalonsod: More coherent naming:
```suggestion
    date_tare = fields.Date(string='Tare date')
```

stock_picking_mgmt_weight/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+    gross_weight = fields.Float(string='Gross')
+    date_gross_weight = fields.Date(string='Gross weight date')
+    net_weight = fields.Float(string='Net', compute="_compute_net_weight")
+
```
dalonsod, line 31: Missing weight unit fields (default=kg, or tons?)

stock_picking_mgmt_weight/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+    def _compute_net_weight(self):
```
dalonsod, line 46: For a computed, non-stored field in v13 every record must be calculated. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 . If not, a cache miss error is thrown

stock_picking_mgmt_weight/views/res_partner_views.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_res_partner_vehicle_form_inherit" model="ir.ui.view">
```
dalonsod, line 3: This view (and others) should only be applied to Vehicles User group

dalonsod (changes requested):
Initial code review, see comments. And,
* `vehicle.vehicle` model can be changed with a button linked to `stock picking`, etc.
* missing `oca_dependencies.txt` file with one initial line (`slv-general https://github.com/solvosci/slv-general`).

ChristianSantamaria:
@dalonsod At the moment the dimensions of the ticket is an A5, it will be necessary to see if it is changed in the future. On the part of the design I followed the original style. What do you think?
Now I have a problem printing the top right code as it does not add top margin, I am working on it. 
imagen

dalonsod:
> 
> 
> @dalonsod At the moment the dimensions of the ticket is an A5, it will be necessary to see if it is changed in the future. On the part of the design I followed the original style. What do you think?
> Now I have a problem printing the top right code as it does not add top margin, I am working on it.
> imagen

My initial comments:
- I like the style. Dimensions should change, but starting from A5 is a good chance.
- By now, don't worry about order number.  We're not sure if that field will be finally printed.
- Regarding to date and hour formatting, I agree using forced formats temporally. The most suitable option should be using a per-user format (depending on his language), but it could be complicated, and not important at this time.
- Regarding to "Kg Weight", the weight conversion to kg should be tons independent. Is it possible? Anyway, if tons are supposed everytime, the right conversion to kg should be multiplying by 1,000, not 100. And, is it possible to truncate decimals and only show integer part?

ChristianSantamaria:
Done, to solve the automatic conversion I have used a function of the uom itself since I saw that it is used in several OCA addons:
https://github.com/odoo/odoo/blob/5ab9a317b1e0b3c1d5a143d49efb04b2f96c6300/addons/uom/models/uom_uom.py#L135
This method also has an automatic rounding but it follows the format of the unit of measure in which it is transformed, so I saw that it can be solved in the report view just by adding this: `<td><t t-esc="'%.0f'% move.get_convert_weight_kg(move.net_weight)"/></td>` https://www.odoo.com/es_ES/forum/ayuda-1/how-can-remove-decimal-number-00-in-print-receipt-150406

dalonsod:
@ChristianSantamaria see improvements for conversion to `kg` function. Looks good to me decimal truncate method in this case.

stock_picking_mgmt_weight/report/stock_picking_tag_report.xml
```diff
@@ -0,0 +1,30 @@
...
+            <field name="margin_left">0</field>
+            <field name="margin_right">0</field>
+            <field name="header_line" eval="False" />
+            <field name="header_spacing">0</field>
+            <field name="dpi">90</field>
+        </record>
+
+        <report id="action_report_picking_tag"
+            model="stock.picking"
+            string="Print Tag (PDF)"
```
dalonsod: Actually this report should be named as follows:
```suggestion
            string="Print Ticket (PDF)"
```

stock_picking_mgmt_weight/report/stock_picking_tag_template.xml
```diff
@@ -0,0 +1,106 @@
...
+    <template id="label_transfer_template_view_pdf">
+        <t t-call="web.basic_layout">
+            <div class="page" style="display: -webkit-box; display: -webkit-flex; display: flex; -webkit-box-pack: center; justify-content: center; align-items: center; -webkit-box-align: center;">
+                <t t-foreach="docs" t-as="picking">
+                    <t t-foreach="picking.move_lines" t-as="move">
+                        <div class="w-100" style="page-break-before: always;">
+                            <div class="w-100 text-right mt-1 mr-1">
+                            </div>
+                            <div class="w-100 text-right mt-1 mr-1">
+                                <span>P21/6986</span>
```
dalonsod: Could you temporary remove this text? But it's still correct the div container (maybe with a comment in order to remember it)

stock_picking_mgmt_weight/report/stock_picking_tag_template.xml
```diff
@@ -0,0 +1,106 @@
...
+                                    <div class="row">
+                                        <div class="w-25">Product:</div>
+                                        <div class="w-75"><t t-esc="move.product_id.name"/></div>
+                                    </div>
+                                    <div class="row">
+                                        <div class="w-25">LoW Code:</div>
+                                        <div class="w-75"><t t-esc="move.product_id.product_tmpl_id.low_code"/></div>
+                                    </div>
+                                    <div class="row mb-3">
+                                        <div class="w-25">Partner:</div>
```
dalonsod: Better use in this case (attending our customer requirment):
```suggestion
                                        <div class="w-25">Company:</div>
```

stock_picking_mgmt_weight/views/stock_picking_views.xml
```diff
@@ -0,0 +1,88 @@
...
+    </record>
+    <!-- New Stock Move Line -->
+
+    <record id="view_stock_picking_weight_form_inherit2" model="ir.ui.view">
+        <field name="name">stock.picking.form (move line)</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//header/button[@name='action_cancel']" position="before">
+                <button name="%(action_report_picking_tag)d" string="Print Tag" type="action" groups="base.group_user"/>
```
dalonsod: This button should be placed in `stock.picking.form (weight)` view, together with the other scale additions, and also only shown if the picking is a scale type one.

dalonsod (changes requested):
See new comments

stock_picking_mgmt_weight/models/stock_move.py
```diff
@@ -0,0 +1,62 @@
+# © 2021 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields, api, _, tools
+from datetime import date, datetime
```
dalonsod: Please confirm if these imports are unnecesary and can be removed:
```suggestion
from odoo import models, fields, api, _
```

dalonsod (changes requested):

stock_picking_mgmt_weight/report/stock_picking_tag_template.xml
ChristianSantamaria: At the moment I have put the content in invisible since if I remove the content and only leave the container, the next container raises it to the initial position... I am looking for a solution to this problem

stock_picking_mgmt_weight/views/stock_picking_views.xml
```diff
@@ -0,0 +1,92 @@
...
+        <field name="name">stock.picking.form (move line)</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//header/button[@name='action_cancel']" position="before">
+                <button string="Open classification "
+                    name="move_weight"
+                    class="btn-primary"
+                    type="object"
+                    attrs="{'invisible': ['|', ('state', 'in', ('done', 'classified', 'cancel'))}"/>
```
dalonsod: This code breaks addon, maybe is incomplete

stock_picking_mgmt_weight/models/stock_picking.py
```diff
@@ -0,0 +1,48 @@
...
+from odoo import models, fields, api, _
+from datetime import date, datetime
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    type_scale = fields.Boolean(compute="_compute_type_scale")
+    vehicle_id = fields.Many2one('vehicle.vehicle', string='Vehicle')
+    state = fields.Selection(selection_add=[('classified', 'Classified')])
```
dalonsod: We've thought about this feature and seems too dangerous, so we'd prefer this approach:
- keep the current picking states.
- create a new `Many2one` field that represents "classification picking"
- when classified, fill the new above field, and cancel the current picking

With these changes, we'll be able to make a custom view for scale weights (those pickings that belongs to picking type with scale flag), and show if are pending of classification, or already classified, and ignore picking state. Finally, these pickings won't affect stock.

stock_picking_mgmt_weight/wizards/move_weight.py
```diff
@@ -0,0 +1,137 @@
...
+                        'product_qty': move_line.product_uom_qty,
+                        'product_uom': move_line.product_uom_id.id,
+                        'price_unit': move_line.product_id.product_tmpl_id.standard_price,
+                        'order_id': purchase_order.id,
+                        'date_planned': datetime.now(),
+                    })
+
+                purchase_order.button_confirm()
+                picking_new = purchase_order.picking_ids
+                picking_new.vehicle_id = self.picking_id.vehicle_id.id
```
dalonsod: Better comment this line, as for classification the vehicle is not important, and pickings will be linked

stock_picking_mgmt_weight/wizards/move_weight.py
```diff
@@ -0,0 +1,137 @@
...
+    def write_operations(self):
+        self.ensure_one()
+        purchase_order = self.env['purchase.order']
+        if self.weight_classified == 0:
+            if not self.move_line_weight_ids.mapped('order_line'):
+
+                order_new = purchase_order.new({
+                    'partner_id': self.picking_id.partner_id.id
+                })
+                order_new.onchange_partner_id()
```
dalonsod: Please encapsulate this code in a method, so other addons could set other fields and fire other methods

dalonsod (changes requested):
Locally tested, see comments

dalonsod:
@ChristianSantamaria I've cleaned commit history.

For an initial beta version IMO only `i18n` and `README` generated files are missing for both addons present in this PR. Could you generate them and squash result in the proper commits?

ChristianSantamaria:
Pending: 
 - add values in classification_order_ids(One2many)
 - origin_purchase_order_id

ChristianSantamaria:
In the end I tried again with this dependency "classification_order_line_ids.qty_received" and it worked correctly

ChristianSantamaria:
New updates:
- Added new button "Add Rest"
- Make the autofill button invisible when there is already a stock move line
- Create a file for stock move line tree view
- Update translations

ChristianSantamaria:
Updates:
  - Added a button to change related order only in classification orders
  - The problem when pressing the button to print ticket pdf was solved
  - The display of the date and time on the ticket is grouped
  - Optimization in the classification wizard
  - Pending: the pending order is not updated when using the change related order button

ChristianSantamaria:
Updates:
- In Vehicles the partner field changed to carrier
- Unlink from contact when filtering and creating vehicles
- Change the position of the field of vehicles in contacts
- Added a default pricelist for purchases
- Fix some integration with the puchase_pricelist addon
- Update translations

ChristianSantamaria:
Updates:
 - In the original order add a button for classification invoices.
 - The order date of a classification order is now equal to the date the ticket was created.

PR #24, opened 2022-01-04
[13.0][IMP] stock_picking_mgmt_weight: several security improvements

dalonsod:
* New maximum weight for scale operations
* `ondelete=restrict` on a lot of `Many2one` fields
* Save classification original user
* Only S/C administrator can change a classification


dalonsod:
cc @ChristianSantamaria

PR #26, opened 2022-01-19
[13.0][ADD] stock_move_action_done_custdate_val

dalonsod:
Passes custom date for a stock move valuation models (valuation layer and account moves) when move is marked as done.

This addon links `stock_account` and `stock_move_action_done_custdate` modules.


dalonsod:
cc @ChristianSantamaria

PR #28, opened 2022-01-21
[13.0][FIX+IMP] stock_picking_mgmt_weight: wrong weight method calculation + some improvements

ChristianSantamaria:
The field with which the total amount was calculated was changed since if another product was added to the purchase order as a discount, the correct weight type was not selected


dalonsod:
@ChristianSantamaria I've finally stored "Announced" vs "Net weight" classification weight mode, and then retrieved when "View classification" is clicked, for more security

PR #30, opened 2022-01-26
[13.0][FIX] stock_picking_mgmt_weight: vehicle users cannot create a scale entry when towing is filled

dalonsod:
With this fix, no new rule is needed


dalonsod:
cc @ChristianSantamaria

PR #31, opened 2022-01-26
[13.0][UPD] stock_picking_report_deliv_vat to stock_picking_partner_vat

ChristianSantamaria:
Addon update, added partner_vat field to form view and renamed addon


stock_picking_partner_vat/models/stock_picking.py
```diff
@@ -0,0 +1,11 @@
...
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    partner_vat = fields.Char(related="partner_id.vat")
+    picking_type_code = fields.Selection(related="picking_type_id.code")
```
dalonsod: I think this field is already present in the main `stock.picking` definition, so it's unnecessary. Could you check

dalonsod (changes requested):

stock_picking_partner_vat/models/stock_picking.py
ChristianSantamaria: Yes, you are right, it exists and with the same name now I make the change, my fault

stock_picking_partner_vat/views/stock_picking_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_stock_picking_form_inherit" model="ir.ui.view">
+        <field name="name">stock.picking (vat)</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"></field>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_id']" position="after">
+                <field name="partner_vat" attrs="{'invisible': [('picking_type_code', 'not in', ('incoming', 'outgoing'))]}"/>
```
dalonsod: I've realized that even an internal transfer could have a partner_id (but optional).
Then, maybe a simpler invisible condition should be this one:
```suggestion
                <field name="partner_vat" attrs="{'invisible': [('partner_vat', '=', False)]}"/>
```
What do you think?

dalonsod (approved):

PR #32, opened 2022-01-28
[13.0][IMP] stock_picking_mgmt_weight

ChristianSantamaria:
Add unit price to tree view sale order line


dalonsod:
Added version bump

PR #35, opened 2022-02-10
[13.0][IMP] stock_picking_mgmt_weight: sale lines menu - new pending column and sum aggregates

dalonsod:
Several improvements for Sale Order Lines menu:

* Added new computed pending quantity field, which is shown in this menu.
* Simplified some quantity column titles.
* Added aggregate sum values for quantity columns.


dalonsod:
cc @ChristianSantamaria

PR #38, opened 2022-04-20
[13.0][IMP] stock_picking_mgmt_weight: pending_qty store in sale order line

ChristianSantamaria:
Change the pending_qty field to `store=True` to be able to filter by it


ChristianSantamaria:
Installed in the test environment, @dalonsod could you test it?

dalonsod (approved):
Tested 👍

PR #39, opened 2022-05-19
[13.0][IMP] stock_picking_mgmt_weight: sale order pending quantities cancellation

dalonsod:
With this improvement, a basic sale order pending quantities cancellation is provided:
* The process is for the whole order, not line per line.
* Pending pickings are cancelled.
* Pending quantity for a line will take in account cancel quantity
* Subsquent order operations (such as adding a new line or increasing line quantity) should lead to inconsistent information.

@ChristianSantamaria could you update test environment with this improvement? Thanks!


ChristianSantamaria:
Done 👍

PR #40, opened 2022-05-25
[13.0][FIX] stock_picking_mgmt_weight: misclassification validation

ChristianSantamaria:
It was only being validated if the outstanding amount was greater than zero. Fix with float_is_zero


dalonsod (approved):
Tested 👍

PR #41, opened 2022-05-31
[13.0][ADD] stock_valuation

ChristianSantamaria:
Add two new models that are the historical day by day and the current one by warehouse both

Currently working with:
  - Purhcase and sale orders
  - Internal movements
  - Purchase and sale returns


ChristianSantamaria:
Now in the stock_valuation_mrp addon the production and unbuild operations are done. 
Soon the commits for both addons will be independent.

dalonsod:
Let's go... merging!!!!

PR #45, opened 2022-06-22
[13.0][IMP] stock_picking_mgmt_weight: improved custom date propagation

dalonsod:
With this change, stock picking custom date propagation to their stock moves will no longer depend on "Validate" button, but to done state setting.
This will prevent corner cases that custom date shouldn't be propagated.

cc @lmiguens-solvos @ChristianSantamaria ASAP I'll update Test environment and this improvement could be properly tested

### DIFF
--- a/stock_picking_mgmt_weight/__manifest__.py
+++ b/stock_picking_mgmt_weight/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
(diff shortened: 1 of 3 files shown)


lmiguens-solvos:
In my tests today I detected that the date of the order is propagated to the delivery note just when Propagate Custom Date is not checked. 
That condition seems to work just in the other way.

lmiguens-solvos:
LGTM

PR #49, opened 2022-09-23
[13.0][IMP] stock_picking_mgmt_weight: enable purchase line cancel pending quants individually

dalonsod:
With this improvement cancelling individual purchase line pending quantities individually becomes possible.

TODO
* Line "cancel pending" button rendering.
* Computed `is_cancellable` field performance with large databases in purchase order lines menu.

cc @ChristianSantamaria


dalonsod:
Self-tested, this is an initial version, ready to be merged!

PR #50, opened 2022-09-26
[13.0][IMP] stock_picking_mgmt_weight: ensure pricelist access for scale/classif users

dalonsod:
With this improvement, scale/classification users can always access pricelists.

This PR prevents problems with solvosci/slv-product-priv#10


dalonsod:
Let's test it in Production environment :-)

PR #51, opened 2022-09-27
[13.0][FIX] stock_picking_mgmt_weight_pot: cancel pending quantities and multi order types

dalonsod:
This fix prevents error when cancelling pending quantities if there are more than one order type available in current classifications. The first one is selected in that case


dalonsod:
cc @ChristianSantamaria

PR #54, opened 2022-10-03
[13.0][FIX] stock_valuation: PHAP obtained from new sale return

dalonsod:
With this fix, when a sale return comes from a sale line splitted in more than one move, an expected singleton when PHAP is calculated is avoided


dalonsod:
cc @ChristianSantamaria

dalonsod:
Already tested in Production environment

PR #56, opened 2022-10-13
[13.0][IMP] stock_picking_mgmt_weight: enable sale line cancel pending quants individually

dalonsod:
With this improvement, sale line pending quantity could be cancelled. Only pending stock moves are cancelled in that case.

cc @ChristianSantamaria @lmiguens-solvos


dalonsod:
Installed in Test environment, @ChristianSantamaria @lmiguens-solvos  could you test? Thanks!

PR #57, opened 2022-10-24
[13.0][FIX] stock_picking_mgmt_weight: SOL cancel pending qty - do not copy qty cancelled

dalonsod:
With this fix, when a sale order is duplicated, cancel quantities of lines aren't propagated.

cc @ChristianSantamaria


dalonsod:
Self-tested, ready to be merged

PR #58, opened 2022-10-31
[13.0][IMP] stock_picking_mgmt_weight: SOL undo cancel operation & better qty cancelled calculation

dalonsod:
With this PR two improvements are achieved:

- "Undo cancel" operation by line is now supported for sale lines.
- Cancelled quantity is now linked to cancelled quantities in related stock moves for each sale line.
- When undo cancel operation is selected, current cancelled stock move is deleted and a new stock move (and stock picking when needed) is created.

The undo cancel operation, when available, is shown with this icon:

imagen

cc @lmiguens-solvos @ChristianSantamaria

### DIFF
--- a/stock_picking_mgmt_weight/__manifest__.py
+++ b/stock_picking_mgmt_weight/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "13.0.1.18.1",
+    "version": "13.0.1.19.0",
     "category": "stock",
     "website": "https://github.com/solvosci/slv-stock",
     "depends": [
--- a/stock_picking_mgmt_weight/i18n/es.po
+++ b/stock_picking_mgmt_weight/i18n/es.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 13.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-13 11:14+0000\n"
-"PO-Revision-Date: 2022-10-13 11:14+0000\n"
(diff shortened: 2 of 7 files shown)


dalonsod:
@lmiguens-solvos @ChristianSantamaria could you Test in customer testing environment? Thanks!

lmiguens-solvos:
I tested this improvements and it seems to work fine, very intuitive, but there is at least one case where it doesn't work well.
Look at the order VXP22/01766 that I canceled the first line and then I undo the cancel operation, but the move line was not generated again.
I can explain it more in detail tomorrow.
I did the order VXP22/01765 also, and in this (later) the undo cancel operation worked well (at the second)....

dalonsod:
Found the problem: undo cancel didn't work as expected when sale order was in locked state (`done`), not Sale order one (`sale`). And in Test environment, when an order is confirmed, it becomes locked as default (later operations should unlock it).

So, I've disabled undo cancel button when an order is locked, you should unlock it in order to undo cancel operation (BTW it seems to be safer and reasoable). Could you test again? :relaxed:

Other option is keep this functionality even if a sale order is locked, temporarily unlocking it during the operation and restoring it once has finished. What do you think?

lmiguens-solvos:
Tested again. Now It seems everything works well. 
And the solution seems correct, I don't think it is necessary to do the extra work if a sale order is locked, temporarily unlocking it during the operation and restoring it once has finished.
The only problem is that they remember (and we remember) that this option is only enabled if the order is unlocked.
The other option would be to always have the "Undo cancel" operation by line icon visible, but give a message that the order must be unlocked y it is locked.

dalonsod:
Customer agrees with this initial approach, so let's merge it and later usage will say whether it's been a good decision or not

PR #60, opened 2022-12-12
[13.0][FIX] stock_picking_mgmt_weight: ensure no cancelled moves when SO confirmation

dalonsod:
For proper cancel and pending quantities calculation, a SO should only have cancelled stock moves that belong to cancel pending quantities operation. With this fix, when a SO is confirmed, every cancelled move is erased to ensure no previous cancelled moves exist, that would break pending and cancelled quantities calculation.

With this fix, when a SO is cancelled, and later confirmed again, previous cancelled quantities by line are lost and they should be individually cancelled again.

@lmiguens-solvos already installed in Test environment, could you test it? Thanks!

### DIFF
--- a/stock_picking_mgmt_weight/__manifest__.py
+++ b/stock_picking_mgmt_weight/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "13.0.1.19.0",
+    "version": "13.0.1.19.1",
     "category": "stock",
     "website": "https://github.com/solvosci/slv-stock",
     "depends": [
@@ -38,6 +38,7 @@
         "views/stock_move_views.xml",
         "wizards/move_line_weight_views.xml",
         "wizards/move_weight_views.xml",
+        "wizards/sale_order_confirm_views.xml",
         "views/stock_picking_type.xml",
         "views/stock_picking_views.xml",
         "views/res_partner_views.xml",
--- a/stock_picking_mgmt_weight/i18n/es.po
+++ b/stock_picking_mgmt_weight/i18n/es.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 13.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-31 09:55+0000\n"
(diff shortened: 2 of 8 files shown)


lmiguens-solvos:
ok, I see that it works as you say.
Although, I only have 2 comments:
1.- even the canceled picks generated by the cancellation of remains are deleted, that would have to be notified to users.
2.- When quantities are changed to less in the order, the cancellation is made by the quantity of the pick, leaving the order quantity <cancelled quantity , but I understand that it is another matter, but I wanted to comment on it.

dalonsod:
> 1.- even the canceled picks generated by the cancellation of remains are deleted, that would have to be notified to users.

What do you propose, raise a warning when a SO is confirmed, if cancelled pickings already exists? There are two different situations: (a) the SO was previously cancelled with no "cancel pending quantities" actions and (b) the SO was previously cancelled with some "cancel pending quantities" actions. Both situations are indistinguishable at this point, because cancelling a SO will create cancelled moves. I mean: any warning message should be general in this case, because we don't know the history of the previous operations (with the current code).

>2.- When quantities are changed to less in the order, the cancellation is made by the quantity of the pick, leaving the order quantity <cancelled quantity , but I understand that it is another matter, but I wanted to comment on it.

Yes, the current process only cancel draft moves, move quantity is not updated. Anyway, if the user undo the operation, pending quantity will be calculated as expected and cancelled set to 0 again.

lmiguens-solvos:
1.-  in fact, the quantities canceled when canceling the order are lost... that's when we no longer differentiate if they are canceled pending or canceled due to canceling the order...
So, I don't know if we can notify something the cancellation of the order if there are quantities canceled... or simply, as you say, give the general warning in the confirmation, and that's all folks
2.- ok, just so we know

dalonsod:
imagen

This message is displayed when main "Confirm" button is clicked and current SO has at least a cancelled stock move linked with it, what do you think?

lmiguens-solvos:
ok, I understand the message and it is correct, but I am going to propose other in case you like them better
" Al confirmar el pedido se borraran todas las cantidades de canceladas existentes actualmente, revise si alguna de ellas era de restos y debe seguir cancelada una vez confirmado el pedido, si es así deben ser canceladas de nuevo después de confirmar el pedido."

PR #61, opened 2022-12-12
[14.0][ADD] stock_secondary_unit_stock_quant

dalonsod:
Adds computed secondary unit quantites to Stock Quant view


dalonsod:
@ChristianSantamaria module finished with pending functionality (quantities in lot views), ready to be tested.

Could you test it and install it in Test Environment, if everything's ok?

PR #65, opened 2023-01-09
[15.0][ADD] stock_picking_report_wo_order_qty

LucPinheiro:
¡¡¡ Welcome @dalonsod !!!


dalonsod (changes requested):
Locally tested, I've found a not-covered case: when a delivery is still undone, ordered column is hidden. In that case, we loose the only information we've got (since nothing has been delivered yet), so in that case ordered quantity should be shown

stock_picking_report_wo_order_qty/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "category": "Stock",
+    "website": "https://github.com/solvosci/slv-stock",
+    "depends": [
+        "stock"
+    ],
+    "data": [
+        "report/report_deliveryslip.xml"
+    ],
+    'installable': True,
+}
```
dalonsod: Add EOF missing empty line

dalonsod (approved):
Tested, 👍 

Please attend comestic requirerd change, and squash commits.

PR #70, opened 2023-02-10
[13.0][IMP] stock_valuation: several visual improvements

dalonsod:
In this commit these new features were added:
* Product template form view, average prices by warehouse are shown if is single-variant.
* Product average price tree view, stock quantity is shown and new zero stock filters are added. Also, new colors depending on stock quantity are used.
* Product average price form view, stock quantity is shown.
* Stock quant view, unit measure for quant is improved, and price is shown.
* Product average price at date, filtered by default for non-zero quant records.


dalonsod:
@ChristianSantamaria could you apply this PR in Test environment?

Some changes could take a long time because of data initialization.

PR #74, opened 2023-03-02
[13.0][FIX] stock_valuation: purchase order line price change to zero coverage

dalonsod:
Before this fix, when a POL price was changed to `0.0`, price recalculation was skipped. Since `0.0` is a price that should be allowed, this fix includes it.

@ChristianSantamaria could you update Test Environment with this fix?

cc @lmiguens-solvos


dalonsod:
Self-tested 👍

PR #77, opened 2023-05-24
[13.0][IMP] stock_picking_mgmt_weight: ticket creation - use net date for numbering

dalonsod:
With this improvement, in report ticket proper date (tare for incoming weights and gross date for outgoing) is printed for general ticket print date.

Also, this date is used for ticket numbering, when date range is used in the correspondent stock picking sequence. This allows to create weight tickets in the past (e.g. last year) with the desired number according this date.

### DIFF
--- a/stock_picking_mgmt_weight/__manifest__.py
+++ b/stock_picking_mgmt_weight/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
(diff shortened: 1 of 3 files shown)


dalonsod:
cc @ChristianSantamaria 

could you install it in AluTest environment and test it? Thanks!

ChristianSantamaria:
Updated and tested 👍

dalonsod:
Approved by customer, let's merge!

PR #78, opened 2023-06-14
[13.0][FIX] stock_picking_mgmt_weight: purchase order line first invoice date - based on accounting date

dalonsod:
Before this fix, for this calculation bill date was used, but accounting date is more accurate.


dalonsod:
cc @ChristianSantamaria

PR #80, opened 2023-06-26
[13.0][IMP] stock_picking_mgmt_weight: purchase notes in classification wizard

dalonsod:
@ChristianSantamaria could you install and test it in AluTest environment?


ChristianSantamaria (approved):
LGTM 👍 

It is compatible with the addon of the images
imagen

PR #81, opened 2023-06-28
[13.0][IMP] stock_picking_mgmt_weight: enable Outgoing Info for internal transfers

dalonsod:
With this improvement, that information is also enabled for internal transfers, when location warehouses aren't the same.


dalonsod:
cc @ChristianSantamaria

PR #82, opened 2023-09-09
[15.0][MIG] stock_move_action_done_custdate

dalonsod:
Standard migration


dalonsod:
Tested across #83 and #84 , let's merge!!

PR #87, opened 2023-10-02
[15.0][FIX] stock_picking_move_description: Fix bug, no description when show lots is active

adriresu:
…ow lots is active


stock_picking_move_description/report/stock_picking_templates.xml
```diff
@@ -32,9 +32,23 @@
...
+        inherit_id="stock.stock_report_delivery_has_serial_move_line"
+    >
+        <xpath
+            expr="//span[@t-field='move_line.product_id']/parent::td"
+            position="after"
+        >
+            <td>
+                <span t-field="move_line.move_id.description_picking"/>
+            </td>
+        </xpath>
```
dalonsod, line 48: This addition should work, please test it:
```suggestion
        </xpath>
        <xpath expr="//span[@t-esc='description']" position="attributes">
            <attribute name="t-if">False</attribute>
        </xpath>        
```

This will prevent description_picking to be repeated, see:

https://github.com/odoo/odoo/blob/5ffafa7e71fc5de4974bcb326b2ea1eb5cb74d5a/addons/stock/report/report_deliveryslip.xml#L219-L224

adriresu:
Updated and squashed.

dalonsod (approved):
Let's merge!

PR #89, opened 2023-10-30
[15.0][ADD] stock_account_replenish_pop

dalonsod:
Fixes problem with rounding adjustment added by stock_account module and pending to be solved.

See https://github.com/odoo/odoo/pull/129800 for more details


dalonsod:
Tested in customer Test environment :+1:

PR #90, opened 2023-11-03
[15.0][IMP] stock_inventory_move_custdate: quantity at date support

dalonsod:
With this improvement, quantity at date is shown and difference could be calculated based on it


dalonsod:
Replaced by #109

PR #91, opened 2023-11-06
[15.0][IMP] stock_picking_move_custdate: custom date management moved to a standalone field

dalonsod:
This new field (Effective Planned Date) will only apply when filled.


dalonsod:
Tested and approved by customer

PR #95, opened 2023-12-22
[13.0][ADD] logistics_planning_base

ChristianSantamaria:
Base addon for logistics management


logistics_planning_base/views/logistics_schedule_menu.xml
```diff
@@ -0,0 +1,54 @@
...
+        id="menu_root"
+        name="Logistics Planning"
+        web_icon="logistics_planning_base,static/description/icon.png"
+    />
+
+    <menuitem
+        id="menu_planning"
+        parent="menu_root"
+        name="Plannings"
+        sequence="10"/>
```
dalonsod, line 107: Please add User group access

logistics_planning_base/views/logistics_schedule_menu.xml
```diff
@@ -0,0 +1,54 @@
...
+        id="menu_output"
+        parent="menu_planning"
+        name="Outputs"
+        action="logistics_schedule_output"
+        sequence="20"/>
+
+    <menuitem
+        id="menu_config"
+        parent="menu_root"
+        name="Configuration"
```
dalonsod, line 126: Please add manager group access

logistics_planning_base/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_logistics_schedule_user,logistics.schedule user,model_logistics_schedule,group_lp_user,1,0,0,0
```
dalonsod: LP User can edit (modify) schedules. Create and delete permissions are correct

logistics_planning_base/security/logistics_planning_base_security.xml
```diff
@@ -0,0 +1,29 @@
...
+            <field 
+                name="implied_ids"
+                eval="[(4, ref('logistics_planning_base.group_lp_user'))]"
+            />
+            <field
+                name="users"
+                eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
+            />
+        </record>
+    </data>
```
dalonsod, line 34: I miss company access rule

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,51 @@
...
+            ("input", "Input"),
+            ("output", "Output"),
+        ]
+    )
+    picking_id = fields.Many2one('stock.picking')
+    stock_move_id = fields.Many2one('stock.move')
+    destination_partner_id = fields.Many2one('res.partner')
+    partner_id = fields.Many2one('res.partner')
+    transport_type = fields.Selection(
+        selection=[
```
dalonsod: Please move selection definition to a python constant (so it could be used by other modules, as we'll need later)

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,51 @@
...
+            ("ground", "Ground"),
+            ("ocean-going", "Ocean-Going"),
+        ]
+    )
+    input_product_id = fields.Many2one('product.product')
+    output_product_id = fields.Many2one('product.product')
+    product_uom = fields.Many2one('uom.uom')
+    product_uom_qty = fields.Float()
```
dalonsod: Add `digits` attribute as usual

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,51 @@
...
+    scheduled_load_date = fields.Date()
+    commitment_date = fields.Date()
+    logistic_price_unit_type = fields.Selection(
+        selection=[
+            ("trip", "Trip"),
+            ("unit", "Unit"),
+        ]
+    )
+    logistic_price_unit = fields.Float()
```
dalonsod: Add `digits` attribute as usual

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,51 @@
...
+    logistic_price_unit_done = fields.Float()
```
dalonsod: Add `digits` attribute as usual

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="logistics_schedule_view_tree" model="ir.ui.view">
+        <field name="name">logistics.schedule.tree</field>
+        <field name="model">logistics.schedule</field>
+        <field name="arch" type="xml">
+            <tree>
```
dalonsod: Make `tree` editable and add a button that enables access to form view

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,96 @@
...
+                <field name="picking_id" />
+                <field name="partner_id" />
```
dalonsod, line 18: Add for both tree and form view `stock_move_id`, with a restricted domain by partner, product, not linked to any logistics schedule yet and in state='done'.

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,96 @@
...
+                <field name="transport_type" />
+                <field name="input_product_id" />
+                <field name="output_product_id" />
+                <field name="product_uom" />
+                <field name="product_uom_qty" />
+                <field name="scheduled_load_date" />
+                <field name="logistic_price_unit_done" />
+                <field name="carrier_id" />
```
dalonsod, line 40: Ensure that every field is shown in tree view that is currently available in .xlsx example file

dalonsod (changes requested):
See initial comments, and also take in account:
- We'll finally remove "input/output" product duality and only preserve a simple `product_id`.
- Add a new `state` field with these initial states: `draft`, `ready`, `done` and `cancel`. We'll later use it determining `readonly` states for some of the fields (other fields will be permanently readonly, but I'll review them later).
- Don't worry about updating translations until we reach a stable version.

dalonsod:
More pending things:
- Logistic user should be read-access granted to `stock.move` and `stock.picking` models, so these should be added to ACLs csv file.
- Add a button box in pickings form that access to existing schedules, and schedule field for stock.move as well.

logistics_planning_base/models/logistics_schedule.py
dalonsod: Still pending this

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,76 @@
...
+    product_uom_qty = fields.Float()
+    scheduled_load_date = fields.Date()
+    commitment_date = fields.Datetime()
+    logistic_price_unit_type = fields.Selection(
+        selection=[
+            ('trip', 'Trip'),
+            ('unit', 'Unit'),
+        ]
+    )
+    logistic_price_unit = fields.Float(digits='Product Unit of Measure')
```
dalonsod: Wrong digits for a currency field (see e.g. https://github.com/odoo/odoo/blob/13.0/addons/purchase/models/purchase.py#L482). In fact, I also miss `currency_id` field

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,107 @@
...
+                <field name="license_plate_1" />
+                <field name="license_plate_2" />
+                <field name="state" readonly="1"/>
+                <field name="note" />
+                <button
+                    name="action_logistics_schedule_form_view"
+                    icon="fa-gear"
+                    class="p-0"
+                    type="object"
```
dalonsod: Is it possible to create an XML action window and call it instead of forcing a `type=object` button?

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,107 @@
...
+                <field name="scheduled_load_date" />
+                <field name="commitment_date" />
+                <field name="logistic_price_unit"/>
+                <field name="logistic_price_unit_done" />
+                <field name="carrier_id" />
+                <field name="license_plate_1" />
+                <field name="license_plate_2" />
+                <field name="state" readonly="1"/>
+                <field name="note" />
+                <button
```
dalonsod: Please add an extra "Cancel" button that will cancel that schedule, only when current state is `draft` or `ready`, accesible for every role (not only for manager one). And create a server action as well, for multi cancel selection,
When a schedule is cancelled we should release `stock_move_id` link, if it was set.

logistics_planning_base/views/stock_picking_views.xml
```diff
@@ -0,0 +1,30 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button name="%(action_stock_picking_logistics_schedule)d" class="oe_stat_button" icon="fa-truck" type="action" groups="logistics_planning_base.group_lp_user">
+                    <div class="o_form_field o_stat_info">
+                        <field name="logistics_schedule_count" widget="statinfo" string="Logistic Schedule"/>
+                    </div>
+                </button>
+            </xpath>
+            <xpath expr="//field[@name='move_ids_without_package']/tree/field[@name='product_uom']" position="after">
+                <field name="logistics_schedule_id" groups="logistics_planning_base.group_lp_user"/>
```
dalonsod: Make it `optional=hide`.

logistics_planning_base/views/stock_picking_views.xml
```diff
@@ -0,0 +1,30 @@
...
+        <field name="res_model">logistics.schedule</field>
+        <field name="view_mode">tree,form</field>
+        <field name="domain">[("picking_id", "=", active_id)]</field>
+    </record>
+
+    <record id="view_picking_form" model="ir.ui.view">
+        <field name="name">stock.picking.form.inherit</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
```
dalonsod, line 17: Instead of using groups for every field, use `groups_id`. I know it will be removed in v16, but...

logistics_planning_base/views/stock_picking_views.xml
```diff
@@ -0,0 +1,30 @@
...
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button name="%(action_stock_picking_logistics_schedule)d" class="oe_stat_button" icon="fa-truck" type="action" groups="logistics_planning_base.group_lp_user">
```
dalonsod: Make this button not visible when there are no schedules yet

logistics_planning_base/views/stock_move_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_move_form" model="ir.ui.view">
+        <field name="name">stock.move.form.inherit</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_move_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='company_id']" position="after">
+                <field name="logistics_schedule_id" groups="logistics_planning_base.group_lp_user"/>
```
dalonsod: Hide this field when unset

ChristianSantamaria:
@dalonsod 
Done, but the comment.. `Is it possible to create an XML action window and call it instead of forcing a type=object button?` I couldn't find another way to send select the res_id from an action, I have been trying to cover it by context or domain using the active_id but the window opened completely empty..

dalonsod:
Ok, I understand. Anyway, try to create XML action. Then, we could keep `type=object` for the button, but within python method retrieve XML action and add `res_id` stuff.

logistics_planning_base/models/logistics_schedule.py
dalonsod: Still pending this: Add `digits` attribute as usual for qty fields

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,110 @@
...
+                    name="action_logistics_schedule_form_view"
+                    icon="fa-gear"
+                    class="p-0"
+                    type="object"
+                    title="Edit"
+                />
+            </tree>
```
dalonsod, line 50: Cancel action should be also available for a single schedule, so add smart button in this view, and in form view as well

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,84 @@
...
+        action = self.env.ref('logistics_planning_base.action_logistics_schedule_form')
+        action.res_id = self.id
+        return action.read()[0]
+
+    def action_logistics_schedule_cancel(self):
+        for record in self.env.context.get("active_ids"):
+            record = self.browse(record)
+            if record.state in ['draft', 'ready']:
+                record.state = 'cancel'
+                record.stock_move_id = False
```
dalonsod: Simpler code IMO (not tested):
```suggestion
        self.browse(self.env.context.get("active_ids", [])).filtered(
            lambda x: x.state in ['draft', 'ready']
        ).write({"state": "cancel", "stock_move_id": False})
```

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def write(self, values):
+        stock_move_id = values.get('stock_move_id')
```
dalonsod: Please make this code multi

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,83 @@
...
+            stock_move_id = values.get('stock_move_id')
+
+            if record.stock_move_id and stock_move_id is not None:
+                record.stock_move_id.logistics_schedule_id = False
+
+            if stock_move_id:
+                move = record.env['stock.move'].browse(stock_move_id)
+                move.logistics_schedule_id = record
+
+            return super().write(values)
```
dalonsod: Bad indentation, so method returns at first `for` iteration

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,117 @@
...
+                <field name="license_plate_1" />
+                <field name="license_plate_2" />
+                <field name="state" readonly="1"/>
+                <field name="note" />
+                <button
+                    name="%(action_logistics_schedule_cancel)d"
+                    icon="fa-ban"
+                    class="p-0"
+                    type="action"
+                    title="Cancel"
```
dalonsod: Do you remember if it was possible adding a `confirm` attribute that enables a warning message before completing such action?

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="logistics_schedule_view_tree" model="ir.ui.view">
+        <field name="name">logistics.schedule.tree</field>
+        <field name="model">logistics.schedule</field>
+        <field name="arch" type="xml">
+            <tree editable="top">
```
dalonsod: Start with tree decoration... `cancel` schedules should be grayed, `done` in green and `draft` in red

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -0,0 +1,117 @@
...
+    <record id="logistics_schedule_view_search" model="ir.ui.view">
+        <field name="name">logistics.schedule.search</field>
+        <field name="model">logistics.schedule</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="picking_id"/>
+                <field name="partner_id"/>
+                <field name="product_id"/>
+                <separator/>
+                <filter string="Ground" name="transport_type_ground" domain="[('transport_type', '=', 'ground')]"/>
```
dalonsod, line 120: Please add a filter for "Pending" schedules (those that are in `draft` or `ready` states). This filter should be enabled by default, so we can exclude those done or cancelled schedules.

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,83 @@
...
+
+TRANSPORT_TYPE = [
+    ('ground', 'Ground'),
+    ('ocean-going', 'Ocean-Going'),
+]
+
+class LogisticsSchedule(models.Model):
+    _name = 'logistics.schedule'
+    _description = 'logistics.schedule'
+
```
dalonsod, line 21: We'll add a default order, .e.g.:
```suggestion
    _order = "scheduled_load_date desc, id desc"
```

logistics_planning_base/models/logistics_schedule.py
```diff
@@ -0,0 +1,83 @@
...
+        default='draft'
+    )
+    type = fields.Selection(
+        selection=[
+            ('input', 'Input'),
+            ('output', 'Output'),
+        ]
+    )
+    stock_move_id = fields.Many2one('stock.move')
+    picking_id = fields.Many2one('stock.picking', related='stock_move_id.picking_id')
```
dalonsod: If this field is non-stored, is it possible to search accross it, or grouping by it?

logistics_planning_base/views/logistics_schedule_views.xml
ChristianSantamaria: I have done several tests and it seems that in the tree view the confirm attribute does not work, instead in the form, for example, i can make a simple wizard to confirm this action, would you prefer?

dalonsod: OK, temporary hide cancel button in `tree` view. And add `confirm` in `form` button view.

logistics_planning_base/models/stock_move.py
```diff
@@ -0,0 +1,10 @@
+# © 2023 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    logistics_schedule_id = fields.Many2one('logistics.schedule')
```
dalonsod: Could you add a new field named `logistics_schedule_disabled`?
- This will be used for stock move domain selection in logistics schedule records, adding it to existing condition.
- By default as `False` (then, record is available).
- Only editable when stock move record is in done state and no `logistics_schedule_id` is linked with it yet. The goal is determine if a stock move is available (or not) for logistics module only when move is finished and is still unused.
- Shown in views for its management.
- Within installation, a hook method (it could be a post init SQL statement) will set every done stock move as `True`, so when this addon is installed, no stock moves could be selected yet (history moves set could be huge and we can avoid it).

logistics_planning_base/views/stock_move_views.xml
```diff
@@ -12,5 +12,16 @@
...
         </field>
     </record>
+    <record id="view_move_form_2" model="ir.ui.view">
+        <field name="name">stock.move.form.inherit</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_move_form"/>
+        <field name="groups_id" eval="[(4, ref('logistics_planning_base.group_lp_manager'))]" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='company_id']" position="after">
+                <field name="logistics_schedule_disabled" attrs="{'readonly': ['|', ('state', '!=', 'ready'), ('logistics_schedule_id', '!=', False)]}"/>
```
dalonsod: Wrong condition (this flag should be available for editing only in `done` state)
```suggestion
                <field name="logistics_schedule_disabled" attrs="{'readonly': ['|', ('state', '!=', 'done'), ('logistics_schedule_id', '!=', False)]}"/>
```

logistics_planning_base/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def post_init_hook(cr, registry):
+    logging.getLogger('odoo.addons.logistics_planning_base').info(
+        'Updating stock.move with logistics_schedule_disabled field to True')
+
+    cr.execute(
+        """
+        UPDATE stock_move
+        SET logistics_schedule_disabled = True
```
dalonsod, line 14: Set it only to True when state is `done`. If not, current draft and pending moves, when later are marked as done, won't be selectable by default. We can only exclude older done moves, that are probably and totally finished, as logistics planning point of view.

dalonsod (changes requested):
See comments. And I also miss new field available in stock move tree within picking form (set it as optional=hide)

dalonsod:
Improvements and fixes in last commit:

- `stock.move` selection: added `logistics_schedule_disabled` condition, name get and name search improvements.
- New `_action_cancel()` standalone method.
- New "Set ready" management. Moved "Mark as done" from `logistics_planning_invoice`.
- Full flow state management (but cancel->draft still pending).
- Readonly states initial management.
- Filter and other form improvements.
- Product quantity dependency from stock move.

dalonsod:
Improvements and fixes in last commit:
- Product Qty `readonly` state fix.
- Schedule renaming (pending to be moved to `name_get()`)
- Stock move re-assignment refactorization.
- Missing settings section menu.
- Logistics schedule in stock move made `readonly`.

dalonsod:
Replaced by #98

PR #96, opened 2024-01-04
[13.0][ADD] logistics_planning_invoicing

ChristianSantamaria:
Adds link with Invoicing

Edit: depends on #95


logistics_planning_invoicing/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+{
+    'name': 'Logistics Planning Invoicing',
+    'summary': '''
+        Adds link with Invoicing
```
dalonsod: Enhance this description:
```suggestion
        Links Logistics Planning with Invoicing
```

logistics_planning_invoicing/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'version': '13.0.1.0.0',
+    'category': 'stock',
+    'website': 'https://github.com/solvosci/slv-stock',
+    'depends': [
+        'account',
+        'logistics_planning_base',
+    ],
+    'data': [
+        'views/account_move_views.xml',
+        'views/logistics_planning_views.xml',
```
dalonsod: This file should be renamed IMO:
```suggestion
        'views/logistics_schedule_views.xml',
```

logistics_planning_invoicing/models/account_move.py
```diff
@@ -0,0 +1,25 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields, api
+
+
+class AccountMove(models.Model):
+    _inherit = 'account.move'
+
+    logistics_schedule_ids = fields.Many2many('logistics.schedule', 'account_move_id')
```
dalonsod: Could this field be `One2many` (a schedule will be only present in an invoice) and `readonly`?

logistics_planning_invoicing/models/account_move_line.py
```diff
@@ -0,0 +1,28 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields
+
+
+class AccountMoveLine(models.Model):
+    _inherit = 'account.move.line'
+
+    logistics_schedule_id = fields.Many2one('logistics.schedule')
```
dalonsod, line 10: This field should be `readonly=True`, because user won't directly fill it (or not at this moment, customer didn't required it). Making it read only, form edition will be disabled.

logistics_planning_invoicing/models/account_move_line.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def create(self, values):
+        res = super().create(values)
+        for record in res.filtered(lambda x: x.logistics_schedule_id):
+            record.logistics_schedule_id.account_move_line_id = record.id
```
dalonsod: aghhhh missing return (without it, this shouldn't work):
```suggestion
            record.logistics_schedule_id.account_move_line_id = record.id
        return res
```

logistics_planning_invoicing/models/account_move_line.py
```diff
@@ -0,0 +1,28 @@
...
+
+from odoo import models, fields
+
+
+class AccountMoveLine(models.Model):
+    _inherit = 'account.move.line'
+
+    logistics_schedule_id = fields.Many2one('logistics.schedule')
+
+    def create(self, values):
```
dalonsod: Missing:
```suggestion
    @api.model
    def create(self, values):
```

logistics_planning_invoicing/models/account_move_line.py
```diff
@@ -0,0 +1,28 @@
...
+            logistics_schedule_id = values.get('logistics_schedule_id')
+
+            if record.logistics_schedule_id and logistics_schedule_id is not None:
+                record.logistics_schedule_id.account_move_line_id = False
+
+            if logistics_schedule_id:
+                ls_id = record.env['logistics.schedule'].browse(logistics_schedule_id)
+                ls_id.account_move_line_id = record
+
+        return super().write(values)
```
dalonsod: This method could be improved, because values is the same for every iteration, so obtaining `values.get('logistics_schedule_id')` could be made once at the beginning of the method.

Indeed, we should skip the whole process if is not passed and consider self as a single record, because we cannot update several account move lines with the same schedule (1 schedule => 1 invoice line), not tested:
```suggestion
        if "logistics_schedule_id" in values:
            logistics_schedule_id = (
                values.get("logistics_schedule_id")
                and self.env["logistics.schedule"].browse(values.get("logistics_schedule_id"))
                or False
            )
            if logistics_schedule_id:
                logistics_schedule_id.account_move_line_id = self
            else:
                self.logistics_schedule_id.account_move_line_id = False
        return super().write(values)
```

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+# License LGPL-3.0 (https://www.gnu.org/licenses/lgpl-3.0.html)
+
+from odoo import models, fields, _
+from odoo.exceptions import ValidationError
+
+
+class LogisticsSchedule(models.Model):
+    _inherit = 'logistics.schedule'
+
+    account_move_line_id = fields.Many2one('account.move.line')
```
dalonsod: This should be `readonly`.

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+    account_move_id = fields.Many2one('account.move', related='account_move_line_id.move_id', string='Invoice')
+
+    def is_invoiceable(self):
+        return any(not record.account_move_line_id or record.state == 'done' for record in self)
+
+    def action_create_invoice_wizard(self):
+        Wizard = self.env['logistics.schedule.account.move.wizard']
+        partner_id = self.env['logistics.schedule'].browse(self.env.context.get("active_ids")).mapped('partner_id')
```
dalonsod: Not partner, but carrier!!
```suggestion
        carrier_id = self.env['logistics.schedule'].browse(self.env.context.get("active_ids")).mapped('carrier_id')
```

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+        if len(partner_id) > 1:
+            raise ValidationError(_("You cannot select multiple records with different partners."))
```
dalonsod: Better this for user:
```suggestion
            raise ValidationError(_("You cannot select multiple schedules with different carriers."))
```

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+        new = Wizard.create({
+            "logistics_schedule_ids": self.env.context.get("active_ids"),
+            'carrier_id': partner_id.id
```
dalonsod: Wrong:
```suggestion
            'carrier_id': carrier_id.id
```

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+            'name': _('Create Invoice'),
+            'res_model': 'logistics.schedule.account.move.wizard',
+            'view_mode': 'form',
+            'view_type': 'form',
+            'res_id': new.id,
+            'target': 'new',
+            'type': 'ir.actions.act_window',
+        }
+
+    def action_logistics_schedule_create_invoice(self):
```
dalonsod: I don't understand this method, with a `return` within the `for` statement. Where is it used?

logistics_planning_invoicing/models/res_config_settings.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import models, fields
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    logistics_schedule_default_journal_id = fields.Many2one(
+        comodel_name='account.journal',
+        related='company_id.logistics_schedule_default_journal_id',
+        string='Default Journal',
```
dalonsod: This description is incomplete (default journal for what?)

logistics_planning_invoicing/models/res_config_settings.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    logistics_schedule_default_journal_id = fields.Many2one(
+        comodel_name='account.journal',
+        domain="[('type', '=', 'purchase')]",
+        string='Default Journal'
```
dalonsod: This label is incomplete

logistics_planning_invoicing/views/account_move_views.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">logistics.schedule</field>
+        <field name="view_mode">tree,form</field>
+        <field name="domain">[("account_move_id", "=", active_id)]</field>
+    </record>
+
+    <record id="view_move_form" model="ir.ui.view">
+        <field name="name">account.move.form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form"/>
```
dalonsod, line 15: Please add security restriction for schedule planning users

logistics_planning_invoicing/views/account_move_views.xml
```diff
@@ -0,0 +1,34 @@
...
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button name="%(action_account_move_logistics_schedule)d" class="oe_stat_button" icon="fa-truck" type="action" attrs="{'invisible': [('logistics_schedule_count', '=', 0)]}">
+                    <div class="o_form_field o_stat_info">
+                        <field name="logistics_schedule_ids" invisible="1"/>
+                        <field name="logistics_schedule_count" widget="statinfo" string="Logistic Schedule"/>
+                    </div>
+                </button>
+            </xpath>
+            <xpath expr="//field[@name='invoice_line_ids']/tree/field[@name='name']" position="after">
+                <field name="logistics_schedule_id" />
```
dalonsod: is this field needed here?

logistics_planning_invoicing/views/account_move_views.xml
```diff
@@ -0,0 +1,34 @@
...
+            </xpath>
+            <xpath expr="//field[@name='line_ids']/tree/field[@name='name']" position="after">
+                <field name="logistics_schedule_id" />
```
dalonsod, line 30: make `optional=hide`.

logistics_planning_invoicing/views/account_move_views.xml
```diff
@@ -0,0 +1,34 @@
...
+            </xpath>
+        </field>
+    </record>
+
```
dalonsod, line 34: I miss schedule field in account move line `form` view.

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+    _inherit = 'logistics.schedule'
+
+    account_move_line_id = fields.Many2one('account.move.line')
+    account_move_id = fields.Many2one('account.move', related='account_move_line_id.move_id', string='Invoice')
+
+    def is_invoiceable(self):
+        return any(not record.account_move_line_id or record.state == 'done' for record in self)
+
+    def action_create_invoice_wizard(self):
+        Wizard = self.env['logistics.schedule.account.move.wizard']
```
dalonsod, line 75: Please restrict access only for Invoicing users (or make server action invisible instead, but I don't know if it's possible)

logistics_planning_invoicing/views/logistics_planning_views.xml
```diff
@@ -0,0 +1,41 @@
...
+        <field name="state">code</field>
+        <field name="code">
+            action = model.action_create_invoice_wizard()
+        </field>
+    </record>
+
+    <record id="logistics_schedule_view_form" model="ir.ui.view">
+        <field name="name">logistics.schedule.form</field>
+        <field name="model">logistics.schedule</field>
+        <field name="inherit_id" ref="logistics_planning_base.logistics_schedule_view_form"/>
```
dalonsod: Add security restriction, only for Invoicing users

logistics_planning_invoicing/wizards/logistics_schedule_account_move_wizard.py
```diff
@@ -0,0 +1,28 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See https://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import fields, models, _
+
+
+class LogisticsScheduleAccountMove(models.TransientModel):
+    _name = 'logistics.schedule.account.move.wizard'
+
+    logistics_schedule_ids = fields.Many2many('logistics.schedule')
```
dalonsod: Make it readonly

logistics_planning_invoicing/wizards/logistics_schedule_account_move_wizard.py
```diff
@@ -0,0 +1,28 @@
...
+    company_id = fields.Many2one('res.company', related='logistics_schedule_ids.company_id')
+    carrier_id = fields.Many2one('res.partner')
```
dalonsod: Make ir readonly

logistics_planning_invoicing/wizards/logistics_schedule_account_move_wizard.py
```diff
@@ -0,0 +1,28 @@
...
+    journal_id = fields.Many2one('account.journal', related='company_id.logistics_schedule_default_journal_id', readonly=False)
```
dalonsod: This is wrong, this field should be independent from default one, only filled by default. With this code, if you change it, you're actually changing the default journal permanently, and such action should be only made from settings.

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+        partner_id = self.env['logistics.schedule'].browse(self.env.context.get("active_ids")).mapped('partner_id')
+        if len(partner_id) > 1:
+            raise ValidationError(_("You cannot select multiple records with different partners."))
+        new = Wizard.create({
```
dalonsod, line 76: Before this, you should check if there are invoiceable schedules selected. And if not, raise an error

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class LogisticsSchedule(models.Model):
+    _inherit = 'logistics.schedule'
+
+    account_move_line_id = fields.Many2one('account.move.line')
+    account_move_id = fields.Many2one('account.move', related='account_move_line_id.move_id', string='Invoice')
+
+    def is_invoiceable(self):
+        return any(not record.account_move_line_id or record.state == 'done' for record in self)
```
dalonsod: Better a single record statement, like a compute field

logistics_planning_invoicing/wizards/logistics_schedule_account_move_wizard.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def create_invoice(self):
+        if self.logistics_schedule_ids.is_invoiceable():
```
dalonsod: This should be unnecessary, because you should load the wizard only if there are invoiceable schedules

dalonsod (changes requested):
Well, some comments 😇

cc @Anxo82

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if not lg_ids.is_invoiceable():
+            raise ValidationError(_("You cannot select records with a done status or with an invoice already assigned."))
+
+        if len(partner_id) > 1:
+            raise ValidationError(_("You cannot select multiple records with different partners."))
+
+        Wizard = self.env['logistics.schedule.account.move.wizard']
+        new = Wizard.create({
+            "logistics_schedule_ids": self.env.context.get("active_ids"),
```
dalonsod: Schedules already invoiced ids should be excluded here

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -18,19 +18,22 @@ def is_invoiceable(self):
         )
 
     def action_create_invoice_wizard(self):
-        lg_ids = self.browse(self.env.context.get("active_ids"))
-        partner_id = lg_ids.mapped('partner_id')
+        if not self.env.user.has_group("account.group_account_user") and not self.env.user.has_group("account.group_account_manager"):
+            raise ValidationError(_('You need invoice user access to perform this action.'))
+
+        lg_ids = self.browse(self.env.context.get('active_ids'))
```
dalonsod: Filter by invoiceable ones

logistics_planning_invoicing/models/account_move_line.py
dalonsod: Use `force_save=1` in view.

dalonsod (changes requested):
See comments

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def action_create_invoice_wizard(self):
+        if not self.env.user.has_group("account.group_account_user") and not self.env.user.has_group("account.group_account_manager"):
+            raise ValidationError(_('You need invoice user access to perform this action.'))
+
+        lg_ids = self.browse(self.env.context.get('active_ids')).filtered(lambda x: x.is_invoiceable)
+        carrier_id = lg_ids.mapped('carrier_id')
+
+        if not lg_ids:
+            raise ValidationError(_('You cannot select records with a done status or with an invoice already assigned.'))
```
dalonsod: This error message is confusing. Try this:
```suggestion
            raise ValidationError(_('No invoiceable schedules were selected'))
```

logistics_planning_invoicing/models/logistics_schedule.py
```diff
@@ -0,0 +1,86 @@
...
+from odoo import models, fields, api, _
+from odoo.exceptions import ValidationError
+
+
+class LogisticsSchedule(models.Model):
+    _inherit = 'logistics.schedule'
+
+    account_move_line_id = fields.Many2one('account.move.line', readonly=True)
+    account_move_id = fields.Many2one('account.move', related='account_move_line_id.move_id', string='Invoice')
+    is_invoiceable = fields.Boolean(compute='_compute_is_invoiceable', store=True, readonly=False)
```
dalonsod: Making editable this field is not a good idea. We've talked about creating another field such `is_finished`, that could be editable when a record is in `ready` state and, at the same time, the record is not invoiceable (we'll have this case later, when link with sales or purchases are developed).
And, when `is_finished` is set to `True`, state must be set to `done` as well.

dalonsod (changes requested):
See comments

logistics_planning_invoicing/models/logistics_schedule.py
dalonsod: > And, when `is_finished` is set to `True`, state must be set to `done` as well.

I still miss this part

ChristianSantamaria: Here https://github.com/solvosci/slv-stock/pull/96/files#diff-d4fc0549398fbf164bb52b4b2d53bae563321fcab352f492e611a95bbb13d67cR17

dalonsod: Sorry, I didn't see it... we'll try to avoid changing fields using a `constrains` method, and use it only for validations. Try to use `write` instead (not create, this field should be readonly during record creation, we'll apply it later).

ChristianSantamaria: Ok, that option is better.

dalonsod:
Changes in last commit:
- Done and cancel new validations.
- Invoice line `create` method move to `api_create_multi`.
- Some view improvements
- Ensure a default journal initially set.
- Marked as `TODO` some pending invoicing features & errors.

dalonsod:
Replaced by #98

PR #100, opened 2024-02-05
[14.0][IMP] stock_barcodes_gs1_secondary_unit: change secondary unit

ChristianSantamaria:


stock_barcodes_gs1_secondary_unit/wizard/stock_barcodes_read_picking.py
```diff
@@ -9,20 +9,19 @@ class WizStockBarcodesReadPicking(models.TransientModel):
...
     def update_fields_after_process_stock(self, moves):
         move_line_id = moves.move_line_ids.filtered(lambda x: x.lot_id == self.lot_id)[0]
         if self.product_qty == move_line_id.qty_done:
             move_line_id.secondary_uom_qty = self.secondary_unit_qty
         else:
             move_line_id.secondary_uom_qty += self.secondary_unit_qty
-        if not move_line_id.secondary_uom_id and move_line_id.product_id.stock_secondary_uom_id:
-            move_line_id.secondary_uom_id = move_line_id.product_id.stock_secondary_uom_id
+        if move_line_id.product_id.production_secondary_uom_id:
+            move_line_id.secondary_uom_id = move_line_id.product_id.production_secondary_uom_id
```
dalonsod: Please hook it:
```suggestion
        new_secondary_uom_id = self._get_new_secondary_uom(move_line_id)
        if new_secondary_uom_id:
            move_line_id.secondary_uom_id = self._get_new_secondary_uom(move_line_id)
```

And then, add a new method that actually should depend on fcd_weight_scale_mrp, but that will work (as in the past did) if it's not installed:

```
def _get_new_secondary_uom(self, move_line_id):
    # production_secondary_uom_id added by "fcd_weight_scale_mrp"
    if "production_secondary_uom_id" in self:
        return move_line_id.production_secondary_uom_id
   else:
       return False
```

dalonsod (changes requested):
See comment, and `fcd_weight_scale_mrp` dependency could be removed (and a bridge method no required)

PR #98, opened 2024-01-31
[13.0][ADD] logistics_planning_*: new addons for logistics management

dalonsod:
Work in progress, already missing purchase/inputs management. This replaces #95 and #96

Edit: depends on #104

PR #104, opened 2024-02-22
[13.0][IMP+FIX] stock_picking_mgmt_weight: added supply condition for purchase lines // prevent picking copy values

dalonsod:
Changes included:
* Added supply condition for purchase lines.
* When creating backorder or picking return, most values or original picking are copied, so they should be defined with `copy=False` if When want to prevent it. With this fix some of them are disabled for copy.


dalonsod:
@ChristianSantamaria required for #98, merging...

PR #106, opened 2024-03-06
[15.0][ADD] stock_orderpoint_prodname

dalonsod:
Makes stock minimum warehouse rule name visible in trees and renames them using product code (or name), so identifying them when used as origin in e.g. a PO is easier.


dalonsod:
An installation hook has been considered, but replaced by this SQL statement:

```
update
	stock_warehouse_orderpoint
set
	name=pp.default_code
from
	stock_warehouse_orderpoint swo
inner join product_product pp
	on pp.id=swo.product_id
where
	stock_warehouse_orderpoint.id=swo.id
	and swo.active=true
	and swo.name like 'OP/%'
	and pp.default_code is not null
;
```

because (1) `OP/` sequence could be previously changed by users and (2) if `default_code` is not available SQL query becomes more difficult.

PR #108, opened 2024-03-25
[16.0][MIG] stock_picking_move_description

Anxo82:
Simple migration to v16


dalonsod:
@Anxo82 show last commit, it worked for me, with

image

when addon is installed the functional-related option is automatically enabled. Please note that it only will take effect if addon _is installed, not updated_, so if you want to see it working, you must uninstall it before! This is right, because such initialization must be placed at a `<data noupdate="1">` section, so admin users feel free to change it, and later addon updates will not force it.

There's only a pending issue: when we have this addon already installed in 15.0 and Odoo is migrated to 16.0... such functionality won't be activated by default... because it's actually an update (from 15.0.x.x.x to 16.0.1.0.0). This could be solved with a migration script that force a `noupudate="1"` section to be executed. We won't make at this time that, because we don't need it in real world (but OCA should require that in their developments).

Well, summarizing:
- Check that it works now, uninstalling and reinstalling it.
- Make pending changes (see next comments)

dalonsod:
@ChristianSantamaria please check functionality group initialization (set) added in last commit, it's a very interesting thing that we've never done before

stock_picking_move_description/security/stock_picking_move_description_security.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="0">
+        <record id="group_picking_report" model="res.groups">
+            <field name="name">Picking report: hide product column</field>
```
dalonsod: Change ID and name, according to actual usage, and better name picking report with official Odoo description (Delivery Slip):
```suggestion
        <record id="group_picking_report_show_prod" model="res.groups">
            <field name="name">Delivery Slip: show product column</field>
```

stock_picking_move_description/models/res_config_settings.py
```diff
@@ -0,0 +1,10 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    group_picking_report = fields.Boolean(implied_group="stock_picking_move_description.group_picking_report")
```
dalonsod: Add a string label with main group title, e.g. the same as new group description

stock_picking_move_description/views/res_config_settings.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='reception_report']" position="after">
+                    <div class="col-12 col-lg-6 o_setting_box" id="group_picking_report">
+                        <div class="o_setting_left_pane">
+                            <field name="group_picking_report"/>
+                        </div>
+                        <div class="o_setting_right_pane">
+                            <label for="group_picking_report"/>
+                            <div class="text-muted">
+                                Picking report: hide product column.
```
dalonsod: Move to a little more detailed description, like
```suggestion
                                When enabled, product column in Delivery Slips will be shown beside line description
```

dalonsod (changes requested):

dalonsod (commented):
LGTM. Please do this administrative suff:
- Add yourself as contributor and enhance addon DESCRIPTION with new feature.
- Squash latest commits with new functionality and documentation above.
- Create another commit with `README.rst` and `index.html` regeneration (`oca-gen-addon-readme` update.

dalonsod (approved):

dalonsod:
@Anxo82 reviewed last commit, you did not this part as I've mentioned:

> Create _**another commit**_ with README.rst and index.html regeneration (oca-gen-addon-readme update.

I see a single commit with the whole functionality (right) but with README.rst and index.html also changed (wrong), those files must be placed in other commit, so if we can port functionality to other version, we prevent also porting that files.

And other thing, in this case don't increase manifest version, we'll only increase it if this improvement was created in an IMP PR made later.

PR #109, opened 2024-03-27
[15.0][FIX+IMP] stock_inventory_move_custdate: partial apply selection bug & quantity at date support

dalonsod:
Replaces #90 

With this improvement, quantity at date is shown and difference could be calculated based on it. Also fixes a partial apply selection bug


dalonsod:
Approved by customer, merging...

PR #116, opened 2024-05-01
[13.0][FIX] logistics_planning_mgmt_weight: add product to planning move filters

dalonsod:
Such filter was missing in stock move custom added domain


dalonsod:
:+1: in AluTest environment

PR #118, opened 2024-05-02
[13.0][FIX] stock_picking_mgmt_weight: show consistent line quantities when editing

dalonsod:
Before this fix, during PO edition, if line quantity changes, other computed stored quantities are temporary shown wrong. This is solved when PO is finally saved. This commit fixes this annoying behavior.

cc @ChristianSantamaria


dalonsod:
:+1:  in AluTest environment

PR #120, opened 2024-05-22
[13.0][IMP+FIX] logistics_planning_[base|sale|mgmt_weight]: some improvements

ChristianSantamaria:


logistics_planning_base/models/logistics_schedule.py
```diff
@@ -158,10 +158,11 @@ def _compute_product_uom_qty(self):
...
         if self.stock_move_id:
-            self.write({
-                "commitment_date": self.stock_move_id.date,
-                "schedule_finished": True,
-            })
+            self.write({"schedule_finished": True})
+            if self.type == 'output':
+                self.write({"commitment_date": self.stock_move_id.date})
+            else:
+                self.write({"effective_date": self.stock_move_id.date})
```
dalonsod: I see it less effective than this implementation (only one `write` access, not tested):
```suggestion
            date_field = "commitment_date" if self.type == 'output' else "effective_date"
            self.write({
                date_field: self.stock_move_id.date,
                "schedule_finished": True,
            })            
```

logistics_planning_sale/models/sale_order_line.py
```diff
@@ -56,7 +60,7 @@ def _prepare_logistics_schedule(self):
             "origin": self.order_id.name,
             "company_id": self.company_id.id,
             "destination_partner_id": self.order_id.warehouse_id.partner_id.id,
-            "partner_id": self.order_partner_id.id,
+            "partner_id": self.order_partner_shipping_id.id,
```
dalonsod: In this case, as partner shipping is not required in any sales line view, I shouldn't create a new field, because it could be simply obtained as follows (like `name` three lines above):
```suggestion
            "partner_id": self.order_id.partner_shipping_id.id,
```

dalonsod (changes requested):
See minor comments. When done, please install int in AluTest Environment. Great work!

dalonsod (approved):
Tested, everything's :+1:

PR #121, opened 2024-05-23
[13.0][IMP] logistics_planning_sale: prevent logistics schedules re-creation when re-confirming a SO

dalonsod:
If a SO with logistics schedules linked is cancelled, set to quotation and confirmed again, logistics schedules are created again. This improvement prevent this, because such LSs should be duplicates of existing ones.

@ChristianSantamaria could you review? Thanks!

### DIFF
--- a/logistics_planning_sale/__manifest__.py
+++ b/logistics_planning_sale/__manifest__.py
@@ -7,7 +7,7 @@
     ''',
     'author': 'Solvos',
     'license': 'LGPL-3',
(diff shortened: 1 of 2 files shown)


ChristianSantamaria:
Tested before and after this commit 
LGTM :+1:

dalonsod:
Ok, we'll install it later in AluTest environment along with other improvements. Let's merge it!

PR #122, opened 2024-05-23
[13.0][IMP] logistics_planning_sale: soft stock move link in sales schedules

dalonsod:
Before this improvement, a stock move can be linked to a logistics schedule that comes from a sales order if sales line matches.

After this improvement, only sale order match is required, so a move that comes from a different sale line (but same sale order) and same product could be linked to LS. This is very useful for sale orders that have several lines with the same product, that generate identic LSs

@ChristianSantamaria could you install it at AluTest environment, so both you and me can test it?


dalonsod:
Tested :+1: let's merge!

PR #124, opened 2024-05-27
[13.0][IMP] logistics_planning_*: manual input sched mark as finished support + multiticket

dalonsod:
_Manual mark as finished_
- With this improvement, new tools for massive schedules mark as finished are added.
- And, when `logistics_planning_mgmt_weight` is installed, schedule inputs are not automatically marked as finished when a ticket (move) is selected, so massive marking schedules as finished becomes important.

_Multi ticket_
- For inputs an "extra tickets" field is added, when more than one tickets need to be linked to a logistics schedule.
- For this new field and also existing ticket field, required product match between logistics schedule and ticket is no longer required.

### DIFF
--- a/logistics_planning_base/__manifest__.py
+++ b/logistics_planning_base/__manifest__.py
@@ -7,7 +7,7 @@
     ''',
     'author': 'Solvos',
     'license': 'LGPL-3',
-    'version': '13.0.1.1.1',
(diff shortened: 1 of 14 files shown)


dalonsod:
:+1: by customer in AluTest environment, let's merge!!

PR #125, opened 2024-05-29
[13.0][IMP] logistics_planning_[purchase|sale|mgmt_weight]: PO/SO and POL/SOL fields updated propagation to LSs support

dalonsod:
With this improvement, those logistics schedules fields that come from purchase/sale order and/or purchase/sale line are automatically updated when origin fields are updated.


dalonsod:
Tested in AluTest environment :+1:

PR #131, opened 2024-06-13
[13.0][FIX] logistics_planning_sale: linking internal stock moves doesn't work

dalonsod:
When added #122 for a soft SO stock move links, internal transfers couldn't be linked any more. This was caused because a `.xml` domain to a related field when desired value is `False` won't work. Making `rel_sale_order_id` field stored fixes it.


dalonsod:
Tested in AluTest environment :+1:

dalonsod:
cc @ChristianSantamaria

PR #133, opened 2024-06-21
[13.0][IMP] logistics_planning_*: allow manual schedules to be duplicated

dalonsod:
With this improvement, it's possible to duplicate those logistics plannings manually created. Some of the fields were reset anyway (like state, linked stock move, invoicing, etc). Parametrized multi duplication is allowed (selecting a certain number of schedules to be duplicated, with a custom limit that prevent a user error selecting "Duplicate" action when it wasn't intended to).


dalonsod:
:+1: in AluTest environment, merging... cc @ChristianSantamaria

PR #134, opened 2024-06-24
[13.0][IMP] logistics_planning_base: allow changing partner for manual schedules

dalonsod:
This is allowed for manual schedules at `ready` state. An extra code that prevents saving a schedule with move but without partner is added


dalonsod:
@ChristianSantamaria :+1: in AluTest environment, I'll merge it

PR #135, opened 2024-06-24
[16.0][ADD] stock_picking_deliver_to

IriaAlonso:
Add deliver_to text field in stock picking and reports


stock_picking_deliver_to/views/stock_picking.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_stock_picking_form" model="ir.ui.view">
+        <field name="name">stock.picking.form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='origin']" position="after">
+                <field name="deliver_to" />
```
dalonsod: This field should be hidden when picking is not an outgoing one (use `picking_type_code` picking field for this)

stock_picking_deliver_to/reports/stock_picking.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <template id="report_stock_picking" inherit_id="stock.report_delivery_document">
+            <xpath expr="//div[@name='div_sched_date']" position="after">
+                <div class="col-auto">
```
dalonsod: Please hide entire structure when `deliver_to` field is empty, so certain space could be saved when not filled

dalonsod (changes requested):
See comments

IriaAlonso:
@dalonsod

dalonsod:
Locally tested, I've filled the new field for an outgoing picking and I didn't see it in delivery slip report. Could you check?

IriaAlonso:
@dalonsod Can you text again?

dalonsod:
Tested again, "Deliver to" now it's shown. But it's shown for every picking, no matter which type is (incoming or outgoing). Please check:

- Code is wrong, you're not referring to current picking document `o.` , si you are actually accessing an undefined variable.
- The exactly condition is "hide report entire structure when _Deliver To_ field is empty" (see below), not when picking is outgoing (that is exactly the opposite condition in some cases, indeed)

IriaAlonso:
@dalonsod can you test again?

dalonsod:
Tested :+1: 

Could you add the following minor improvements:
* For picking tree view, add deliver to field, as `optional="hide"` by default.
* For picking search view, add deliver to field as searchable

IriaAlonso:
@dalonsod

dalonsod (approved):

PR #137, opened 2024-06-28
[16.0][IMP] stock_picking_delivery_slip_a5: change font size

IriaAlonso:
change font size of A5 report


IriaAlonso:
Before:
imagen
After:
imagen

dalonsod:
Cool!

Anyway, I see more changes than simply reducing `font-size`, could you explain why those changes should be needed?

And, at last, for `page` class, which was original rendered font size?

IriaAlonso:
imagen
I think there is too much space between the rows like this, what you think? @dalonsod

IriaAlonso:
new max capacity
imagen

dalonsod (approved):
Tested by customer :+1: let's merge!

PR #138, opened 2024-06-28
[16.0][ADD] stock_picking_deliver_to

IriaAlonso:
Add deliver_to text field in stock picking and reports


dalonsod:
What's this PR? It seems a duplicate of already merged #135

PR #139, opened 2024-06-28
[13.0][FIX] logistics_planning_base: partner readonly check fails for readonly users

dalonsod:
If a user that have read-only access to logistics schedules loads any view that computes `partner_readonly` technical field, an error is raised.

Addon `logistics_planning_mgmt_weight` adds a menu for this limited access users, so an error is raised. This fixes it.

### DIFF
--- a/logistics_planning_base/__manifest__.py
+++ b/logistics_planning_base/__manifest__.py
@@ -7,7 +7,7 @@
     ''',
     'author': 'Solvos',
     'license': 'LGPL-3',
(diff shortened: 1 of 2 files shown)


dalonsod:
cc @ChristianSantamaria computed non-stored fields, when are read, work as a `write()` operation, so a read-only access user, when tries to read such field in a record, is actually trying to updating it, so an error is raised. This could be prevented using `compute_sudo=True`.

PR #141, opened 2024-07-08
[13.0][IMP] logistics_planning_* and stock_picking_mgmt_weight: improvements around Vehicle Type and and Container Number

ChristianSantamaria:
Some improvements:
- New field container_number that is relation with picking field
- New field vehicle_type
- New field etd_date only output and ocean-going


logistics_planning_base/models/logistics_schedule.py
```diff
@@ -158,6 +159,15 @@ class LogisticsSchedule(models.Model):
         Tecnical field indicating whether this record can be set to done state
         """,
     )
+    etd_date = fields.Date(
+        string="ETD",
```
dalonsod: Keep better complete field name in definition, and put `string="ETD"` only in XML views.
```suggestion
        string="ETD Date",
```

logistics_planning_base/views/logistics_schedule_views.xml
```diff
@@ -404,6 +409,9 @@
             <field name="partner_id" position="attributes">
                 <attribute name="string">Customer</attribute>
             </field>
+            <field name="transport_type" position="after">
+                <field name="etd_date" optional="show" attrs="{'invisible': [('transport_type', '!=', 'ocean-going')]}"/>
```
dalonsod: In a `form` view optional attribute is unused:
```suggestion
                <field name="etd_date" attrs="{'invisible': [('transport_type', '!=', 'ocean-going')]}"/>
```

logistics_planning_purchase/models/purchase_order.py
```diff
@@ -45,6 +45,9 @@ class PurchaseOrder(models.Model):
     ls_transport_type = fields.Selection(
         related="incoterm_id.ls_sale_transport_type",
     )
+    logistics_vehicle_type_id = fields.Many2one(
```
dalonsod: As described in current requirements, at this time we won't adapt this field for purchases

logistics_planning_sale/models/sale_order_line.py
```diff
@@ -39,6 +39,14 @@ class SaleOrderLine(models.Model):
         are allowed for this line
         """,
     )
+    logistics_vehicle_type_id = fields.Many2one(
```
dalonsod: This field is uneeded, as `stock_picking_mgmt_weight` already adds it (see https://github.com/solvosci/slv-stock/blob/8b8dea9b8bb6ac0648c33eda7d989aec20b0c3de/stock_picking_mgmt_weight/models/sale_order_line.py#L36)

logistics_planning_sale/models/sale_order_line.py
```diff
@@ -39,6 +39,14 @@ class SaleOrderLine(models.Model):
...
+        'vehicle.type',
+    )
+
+    @api.onchange('logistics_vehicle_type_id')
+    def _onchange_logistics_vehicle_type_id(self):
+        if self.logistics_vehicle_type_id:
+            self.ls_transport_type = self.logistics_vehicle_type_id.ls_sale_transport_type
```
dalonsod: This code (adapted to existing field), should be moved to `logistics_planning_mgmt_weight` due to needed dependencies.

dalonsod (changes requested):
Only code review, see comments

ChristianSantamaria:
Done :+1:

logistics_planning_base/models/logistics_schedule.py
dalonsod: Still missing the second part (and put string="ETD" only in XML views.)

logistics_planning_sale/models/sale_order_line.py
dalonsod: This code is been lost, isn't it?

logistics_planning_base/models/logistics_schedule.py
ChristianSantamaria: Only in tree view?

logistics_planning_sale/models/sale_order_line.py
ChristianSantamaria: This code already existed
https://github.com/solvosci/slv-stock/blob/9d9452d9a94f0f030d2b89a4c77c287d6ad11601/logistics_planning_mgmt_weight/models/sale_order_line.py#L10-L13

logistics_planning_base/models/logistics_schedule.py
dalonsod: In both tree & form. Original complete field name will be only available when inspecting fields model structure

dalonsod:
Validated by customer

PR #142, opened 2024-07-09
[13.0][FIX] stock_picking_mgmt_weight: key fields without `copy=False`

dalonsod:
With this fix incorrect links between classifications and original purchase orders are prevented when duplicating existing and completed purchase orders.


dalonsod:
@ChristianSantamaria merging this...

PR #144, opened 2024-07-24
[13.0][IMP] logistics_planning_invoicing: grouping schedules in the same invoice line

dalonsod:
This improvement enables by default grouping selected logistics schedules to be invoiced in the same line if some of them shares the same values, making aggregation according price unit type expected behavior.

Grouping criteria by:
- product (replacing account, that is not available at this point)
- analytic account
- product unit of measure
- price unit type
- (OPT) price unit

When aggregating,
- If product unit of measure is 'trip' => adding to price unit
- If product unit of measure is 'unit' => adding to price qty

-------

This PR also adds a minor improvement for LP forms (added icon for "Create Invoice" action)

### DIFF
--- a/logistics_planning_invoicing/__manifest__.py
+++ b/logistics_planning_invoicing/__manifest__.py
@@ -7,7 +7,7 @@
     ''',
     'author': 'Solvos',
     'license': 'LGPL-3',
-    'version': '13.0.1.3.0',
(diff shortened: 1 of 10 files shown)


dalonsod:
Tested and approved by customer, ready to be merged!

PR #147, opened 2024-09-13
[13.0][FIX] logistics_planning_invoicing: duplicate, delete and schedule count issues

dalonsod:
This commit fixes the following bugs:
* When an invoice was duplicated, related schedules were also linked to the new invoice. A schedule cannot be linked to more than an invoice line.
* When a draft invoice with linked schedules was deleted, every schedule was unlinked as well (ok), but the same didn't occur when a single invoice line was removed. Now, such check is made at invoice line level.
* Invoice schedule count became wrong when lines with linked schedules were remove.

cc @ChristianSantamaria

### DIFF
--- a/logistics_planning_invoicing/__manifest__.py
+++ b/logistics_planning_invoicing/__manifest__.py
@@ -7,7 +7,7 @@
     ''',
     'author': 'Solvos',
     'license': 'LGPL-3',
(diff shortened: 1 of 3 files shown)


dalonsod:
:+1: in AluTest environment, ready to be merged!

PR #152, opened 2024-10-11
[17.0][MIG] stock_picking_report_wo_order_qty

IriaAlonso:
@dalonsod


dalonsod:
Tested, I see a problem when activating this report option:

imagen

With this option, a new optional column (for lots and serial numbers) is added to the report. In this case, printing a Delivery Note is shown as below:

imagen

The goal is showing done pickings with non-tracked products like this:

imagen


and for lot or s/n tracked products like this:

imagen


Could you check?

IriaAlonso:
@dalonsod change done

dalonsod (approved):
Tested :+1: 

Added @IriaAlonso as contributor, let's merge!

PR #154, opened 2024-11-22
[13.0][IMP] logistics_planning_purchase: adds default product for one line

ChristianSantamaria:
@dalonsod 
In the case of prepare schedule for a line, instead of selecting the product of the line, the default product for purchases of a line is selected in res.config.settings


dalonsod (changes requested):
For default product, please restrict it to storable products.

dalonsod:
Could you install it in AluTest environment?

dalonsod (approved):
Tested :+1:

PR #157, opened 2024-12-05
[15.0][ADD] stock_product_sn_reservation

dalonsod:
Enables s/n secure reservation in moves, preventing undesired situations when a s/n is manually assigned.

This is initially applied to outgoing stock pickings, but it can me also implemented for e.g. MRP productions


dalonsod:
Approved by customer, let's go!!

PR #158, opened 2024-12-11
[17.0][ADD] stock_picking_backorder_partner_config

IriaAlonso:
- Generate partial delivery/receipt according to customer/supplier:
   - create two selections `Partial receipt creation policy` and `Partial delivery creation policy`:
      - `Default`: to behave like the standard (which asks when appropriate).
      - `Ask` --> update translations
      - `Never create`
      - `Always create`
    
@dalonsod


IriaAlonso:
`never creation` opcion --> change `skip_backorder` context to `True`

res_partner_stock_picking/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
+# © 2024 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import models
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _pre_action_done_hook(self):
+        for picking in self:
+            never_parcial_recepcion = picking.partner_id.create_partial_reception
```
dalonsod: Calling this variable "never" is not accurate... it's simply partial reception value for picking contact, we don't know its value at this point. And please check typos for "parcial" and "recepcion" (at least they are mistakes, in english,...)

res_partner_stock_picking/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
...
+            if never_parcial_recepcion == 'never':
+                self.with_context(skip_backorder=True)
+            else:
+                super()._pre_action_done_hook()
```
dalonsod: This code is wrong because of two problems:
* For `never` case `super()` method is never called. You only change context, and then exit.
* When else clause is executed you're calling original method for every picking in self, not for the current picking in the loop, because `super()` is a shortcut of `super(StockPicking, self)`, e.g. called for the whole records.

dalonsod (changes requested):
I'm sure this is not still tested, but reviewed anyway

IriaAlonso:
@dalonsod can you review it?

DantePereyra:
Tested, it works good!

dalonsod (changes requested):
@IriaAlonso please check, both reception and delivery policies options for partners are created, but only reception one used. Depending on picking type, it should be applied the proper one.

Addon name... `res_partner_stock_picking` doesn't rely on final purpose of this: becoming "create backorder" policies dependent on partner configuration. Could you propose another candidate name?

IriaAlonso:
> Addon name... `res_partner_stock_picking` doesn't rely on final purpose of this: becoming "create backorder" policies dependent on partner configuration. Could you propose another candidate name?

Maybe, it could be `stock_picking_backorder_res_partner` or `stock_picking_backorder_partner_config`, what do you think?

dalonsod:
`stock_picking_backorder_partner_config` sound good!

stock_picking_backorder_partner_config/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+{
+    "name": "Stock Picking Backorder Partner Config",
+    "summary": """
+        Generate partial delivery/receipt according to customer/supplier configuration
+    """,
+    "author": "Solvos",
+    "license": "LGPL-3",
+    "version": "17.0.1.0.0",
+    "category": "Sales/Sales",
```
dalonsod: This category is too specific, better use one relating to Inventory

stock_picking_backorder_partner_config/models/res_partner.py
```diff
@@ -0,0 +1,26 @@
...
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    create_partial_reception = fields.Selection(
+        [('by_default', 'By default'), ('ask', 'Ask'), ('always', 'Always create'), ('never', 'Never create')],
```
dalonsod: This could be moved to a Python variable and then used for this field and the below one.

stock_picking_backorder_partner_config/models/stock_picking.py
```diff
@@ -0,0 +1,59 @@
...
+
+        backorder_pickings = super()._check_backorder()
+        for picking in self:
+            if create_partial_picking != 'ask':
+                continue
+            backorder_pickings |= picking
+        return backorder_pickings
+
+    def button_validate(self):
+        create_partial_picking = self.get_create_partial_picking()
```
dalonsod, line 42: `button_validate` is a multi method (see https://github.com/odoo/odoo/blob/f9726cfe93e8850a38d9de06acfa5d78473b50b0/addons/stock/models/stock_picking.py#L1134), so this code should be re-factorized

stock_picking_backorder_partner_config/models/stock_picking.py
```diff
@@ -0,0 +1,59 @@
...
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _pre_action_done_hook(self):
+        if not self.env.context.get('skip_partner'):
```
dalonsod: Using `get` is a good point, but I miss the desired default value (the second parameter), that helps us to check which is the default behavior.

stock_picking_backorder_partner_config/views/res_partner_view.xml
```diff
@@ -0,0 +1,15 @@
+<odoo>
+
+    <record id="res_partner_form_view" model="ir.ui.view">
+        <field name="name">res.partner.form.view (stock_picking_backorder_partner_config)</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='category_id']" position="after">
```
dalonsod: There's an Inventory section under Sales & Purchase page, this could be a better location for theses settings

stock_picking_backorder_partner_config/views/res_partner_view.xml
```diff
@@ -0,0 +1,15 @@
...
+                <field name="create_partial_reception" />
+                <field name="create_partial_delivery" />
```
dalonsod: Make fields required, as we always are expecting a value (and have a default value as well)

dalonsod (changes requested):
Only code review, see comments

IriaAlonso:
@dalonsod changes done

stock_picking_backorder_partner_config/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    partial_policy_selection = [
+        ('by_default', 'By default'),
+        ('ask', 'Ask'),
+        ('always', 'Always create'),
```
DantePereyra, line 9: Could you put `Always` as default?

DantePereyra (changes requested):
LM new requirements on ticket SLV-HT00390

IriaAlonso:
@DantePereyra `always` by default done

DantePereyra:
There is a new issue: when `Ask` is selected, the partial delivery wizard still appears even after all items have been received or delivered. If there are no remaining items to process, this shouldn’t happen. Could you please check?

IriaAlonso:
Done

stock_picking_backorder_partner_config/models/stock_picking.py
dalonsod: This comment isn't still attended. We could check it together @DantePereyra @IriaAlonso

stock_picking_backorder_partner_config/models/stock_picking.py
```diff
@@ -0,0 +1,78 @@
...
+from odoo import models
+from odoo.tools.float_utils import float_is_zero, float_compare
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _pre_action_done_hook(self):
+        if not self.env.context.get('skip_partner', False):
+            for picking in self:
+                create_partial_picking = self.get_create_partial_picking()
```
dalonsod: Ouch, incorrect self reference within a for statement:
```suggestion
                create_partial_picking = picking.get_create_partial_picking()
```

stock_picking_backorder_partner_config/models/stock_picking.py
```diff
@@ -0,0 +1,78 @@
...
+
+                if create_partial_picking in ('by_default', 'ask'):
+                    return super(StockPicking, picking)._pre_action_done_hook()
+                else:
+                    skip_backorder = picking.with_context(skip_backorder=True)
+                    return super(StockPicking, skip_backorder)._pre_action_done_hook()
+        return True
+
+    def _check_backorder(self):
```
dalonsod, line 22: This method is also `multi` and should be refactorized, like `button_validate`.

stock_picking_backorder_partner_config/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
+# © 2025 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    partial_policy_selection = [
```
dalonsod: This works, but a more accurate code should be place it as a external variable, outisde class definition, like e.g. https://github.com/OCA/timesheet/blob/2db1202801bcfc818945b367b4a4f698f2d5811a/hr_timesheet_day_week/models/account_analytic_line.py#L6

Use caps in order to diffenciate this, as a known convention

PR #159, opened 2025-01-02
[13.0][ADD] stock_move_lot_qty_total

dalonsod:
Enables configuring desired products tracked by lot to be delivered using current available stock.

This is achieved enabling for a selected, lot tracked product, the *Force available quantity in deliveries* check. Then, for an outgoing picking line in a pending state, its possible to manually select destination lots and, then, clicking in fill available stock for already filled lots.

When a picking is validated a current stock for lots is checked again, preventing mark as done if a lot is filled with a done quantity different than available one.


dalonsod:
:+1:  in AluTest and approved by customer, at least at this first version

PR #160, opened 2025-01-16
[13.0][FIX] stock_move_lot_qty_total: missing rounding in served quantity comparison

dalonsod:
Strict comparison between two float fields should be implemented using `float_compare` if possible. Quantities in a stock move line and a stock quant could differ in a n-th decimal and break comparison. This fixes it.


dalonsod:
OK in AluProd

PR #161, opened 2025-01-20
[13.0][ADD] stock_picking_type_show_operations_keep

dalonsod:
Preserves Show Operations option in Picking Type, that could be forced to
`true` sometimes across `res.config.settings`.

This is due to standard `stock` addon behavior when lot, warehouse and location
options are set together.

This addon covers a bug already fixed in stock `addon`.
    
See:
* https://github.com/odoo/odoo/commit/d54a58706225fdfc920a03064b750769d75d91e0
* https://github.com/odoo/odoo/commit/6f7e9a93048155e403e6cc56d24595c281297124
    
This addon is intended to cover such bug when Odoo is placed in an earlier version and is not updatable, and shouldn't be used elsewhere.


dalonsod:
cc @ChristianSantamaria

dalonsod:
Not fully tested (cannot reproduce in AluTest original bug), but will install it in AluProd anyway 💪

PR #162, opened 2025-01-24
[17.0][ADD] stock_picking_state_to_process

DantePereyra:
Add a link to show orders in "assigned" or "confirmed" status in "to process".
Add a days field in stock.picking.type, so that when the delay is greater than that, it appears as "very late".

#HT00414

*Edited to change the description according to the functionalities


DantePereyra:
@ChristianSantamaria can you review it?

DantePereyra:
Added the translations with label changes to show new states requiered.
BEFORE:
imagen

AFTER:
imagen

ChristianSantamaria:
The second commit is very tailored for the client, I think it is better to move it to a private repository or to slv-reports and add it to reports_scl... how do you see it @dalonsod ?

stock_picking_state_to_process/models/stock_picking.py
```diff
@@ -0,0 +1,54 @@
...
+
+    scheduled_date_very_late = fields.Datetime(compute='_compute_scheduled_date_very_late', store=True)
+    state = fields.Selection([
+        ('draft', 'Draft'),
+        ('waiting', 'Waiting Another Operation'),
+        ('confirmed', 'Waiting'),
+        ('assigned', 'Ready'),
+        ('done', 'Done'),
+        ('cancel', 'Cancelled'),
+    ])
```
dalonsod, line 20: I see, you've simply rewritten state selection definition (in the exactly same way than Odoo does), and it enables us to get and edit translations, doesn't it?

stock_picking_state_to_process/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock Picking State To Process",
+    "summary": """Changes the state of the picking to 'To process (Stock Pending)' when is 'waiting' and 'To process' when is 'assigned' or 'confirmed'.
+The button in 'Delivery Orders' shows all 'To Process' pickings.""",
+    "version": "17.0.1.0.0",
+    "author": "Solvos",
+    "category": "stock",
+    "license": "LGPL-3",
+    "website": "https://github.com/solvosci/slv-sale",
```
ChristianSantamaria: https://github.com/solvosci/slv-stock

stock_picking_state_to_process/models/stock_picking.py
```diff
@@ -0,0 +1,46 @@
...
+            count = {picking_type.id: count for picking_type, count in data}
+            for record in self:
+                record[field_name] = count.get(record.id, 0)
+
+    def get_action_picking_tree_to_process(self):
+        return self._get_action('stock_picking_state_to_process.action_picking_tree_to_process')
+
+    def get_action_picking_tree_very_late(self):
+        return self._get_action('stock_picking_state_to_process.action_picking_tree_very_late')
+
```
ChristianSantamaria: EOF end line

stock_picking_state_to_process/models/stock_picking.py
```diff
@@ -0,0 +1,46 @@
+# © 2025 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+import time
```
ChristianSantamaria, line 3: Is this dependency necessary if we already have timedelta y date?

stock_picking_state_to_process/models/stock_picking.py
DantePereyra: Yes, that way we can manipulate the labels without compromising the logic

ChristianSantamaria (changes requested):
Some improvements

stock_picking_state_to_process/models/stock_picking.py
DantePereyra: Yes, it´s used in this domain:
`'count_picking_very_late':  [('scheduled_date_very_late', '<', time.strftime(DEFAULT_SERVER_DATETIME_FORMAT)), ('state', 'in', ('assigned', 'waiting', 'confirmed'))]`
In the line 37

DantePereyra:
Improvements done

PR #163, opened 2025-01-25
[13.0][IMP] logistics_planning_sale: enable sale line initial schedule count to be copied

dalonsod:
With this improvement, when a sale line is copied, initial schedule count is copied as well, that is useful for users.


dalonsod:
Tested in AluTest environmet :+1: 

cc @ChristianSantamaria

PR #165, opened 2025-01-31
[13.0][FIX] stock_picking_mgmt_weight: purchase lines - prevent copy incorrect fields

dalonsod:
These three fields were copied during a PO duplication, their content is unique for every purchase line and shouldn't be copied.

cc @ChristianSantamaria 

HT00337


dalonsod:
:+1:  in AluTest, let's merge!

PR #166, opened 2025-01-31
[13.0][IMP] logistics_planning_purchase: enable purchase line initial schedule count to be copied

dalonsod:
With this improvement, when a purchase line is copied, initial schedule count is copied as well, that is useful for users.

cc @ChristianSantamaria 

Required for HT00537


dalonsod:
:+1: in AluTest, let's go!

PR #167, opened 2025-02-06
[13.0][FIX] stock_move_lot_qty_total: prevent duplicated lot selection

dalonsod:
Before this fix, it was possible to select twice or more a lot during lot selection.

With this fix, when smart lot quantity selection is used, lot duplication raises an error.


dalonsod:
ready to be merged

cc @ChristianSantamaria

PR #172, opened 2025-05-05
[17.0][ADD] stock_move_no_merge_moves_from_sale_orders

IriaAlonso:
Prevent the merging of stock moves from sale orders.
HT00718
@dalonsod


stock_move_no_merge_moves_from_sale_orders/models/sale_order_line.py
```diff
@@ -0,0 +1,11 @@
...
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    def _action_launch_stock_rule(self, previous_product_uom_qty=False):
+        return super(SaleOrderLine, self.with_context(from_sale_order=True))._action_launch_stock_rule(previous_product_uom_qty)
```
dalonsod: This is correct, but I'd prefer call by name the parameter:
```suggestion
        return super(SaleOrderLine, self.with_context(from_sale_order=True))._action_launch_stock_rule(previous_product_uom_qty=previous_product_uom_qty)
```

stock_move_no_merge_moves_from_sale_orders/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# © 2025 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License LGPL-3 - See https://www.gnu.org/licenses/lgpl-3.0.html
+{
+    "name": "Stock Move No Merge Moves From Sale Orders",
+    "summary": """
+        Prevent the merging of stock moves from sale orders.
+    """,
+    "version": "17.0.1.0.0",
+    "license": "LGPL-3",
+    "category": "Human Resources",
```
dalonsod: Fill the proper category

dalonsod (changes requested):
See minor comments and we'll test it together

PR #174, opened 2025-05-08
[15.0][IMP] stock_product_sn_reservation: support for internal transfers

dalonsod:
Before this improvement, internal transfers could bypass S/N reservation lock. After this, is treated like outgoing pickings.


dalonsod:
Tested in CtmTest environment :+1:

PR #175, opened 2025-05-19
[17.0][ADD] stock_picking_batch_delivery_carrier

DantePereyra:
Adds Delivery Carrier field on header of stock.picking.batch views, and filters picking_ids by it.


dalonsod:
OK, @ChristianSantamaria could you make this installable it in CmhProd environment? Once done, I'll install it along with @DantePereyra

dalonsod:
Edit: Slv ticket HT00756

ChristianSantamaria:
Done :+1: 
Prepared to install
imagen

DantePereyra:
New changes done.
Prevent carrier changes on pickings already assigned to a batch.

Updated batch creation from stock.picking tree view:
- Context now includes the common carrier.
- Creation is blocked if selected pickings have different carriers.

stock_picking_batch_delivery_carrier/wizard/stock_picking_to_batch.py
```diff
@@ -0,0 +1,17 @@
...
+class StockPickingToBatch(models.TransientModel):
+    _inherit = 'stock.picking.to.batch'
+
+    def attach_pickings(self):
+        pickings = self.env['stock.picking'].browse(self.env.context.get('active_ids'))
+        carriers = pickings.mapped('carrier_id')
+        if len(carriers) > 1:
+            raise UserError(_("All selected pickings must have the same carrier."))
+        carrier_id = carriers.id if carriers else False
+        self = self.with_context(default_carrier_id=carrier_id)
```
dalonsod: Are you covering these cases?
* There's no carrier for the batch => an error should be raised, because carrier is now mandatory for batches.
* Some pickins belong to a carrier (the same one), but at least another one has no carrier => an error should be raised, because every picking should have a carrier assigned

stock_picking_batch_delivery_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,17 @@
...
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+from odoo import models, api, _
+from odoo.exceptions import UserError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.constrains('carrier_id')
+    def _check_carrier_id(self):
```
ChristianSantamaria, line 11: @dalonsod Wouldn't using the write method be better than using constraints?

stock_picking_batch_delivery_carrier/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import stock_picking_batch
+from . import stock_picking
```
ChristianSantamaria: EOF Line :/

ChristianSantamaria (changes requested):
Some minor changes

stock_picking_batch_delivery_carrier/models/stock_picking.py
dalonsod: if we strictly need to check a value (pass it if it's right, or raise an error otherwise), `api.constrains` is the simplest option. `write` is required e.g. if we also need to update another fields (we shouldn't perform create/update operations within a constrains method).

Thus, for this case `constrains` seems to be fine.

Anyway @DantePereyra you could move `for`-`if` statement in a `for`-`filtered` one, and raising an error for all pickings involved together.

DantePereyra:
Changes done! @ChristianSantamaria

PR #179, opened 2025-07-03
[17.0][ADD] product_stock_all_companies

IriaAlonso:
Show the stock of a product in all the warehouses, even if they are from other companies
ticket HT00470
@dalonsod


IriaAlonso:
show product stock on template when it hasn't more than one variant
@dalonsod

product_stock_all_companies/models/product_template.py
```diff
@@ -0,0 +1,23 @@
...
+    stock_by_company_ids = fields.Many2many(
+        'stock.company',
+        compute='_compute_stock_by_company_template',
+    )
+    count_attribute_line_ids = fields.Float(compute='_compute_stock_by_company_template')
+
+    @api.depends('product_variant_ids.stock_by_company_ids', 'attribute_line_ids')
+    def _compute_stock_by_company_template(self):
+        for template in self:
+            template.count_attribute_line_ids = len(template.attribute_line_ids)
```
dalonsod: You can simply use already existing `product_variant_count` template field, and `count_attribute_line_ids` becomes unnecessary

product_stock_all_companies/views/product_template_views.xml
```diff
@@ -0,0 +1,20 @@
+<odoo>
+    <record id="product_normal_form_view" model="ir.ui.view">
+        <field name="name">product.template.form </field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="stock.view_template_property_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='inventory']//field[@name='description_picking']/.." position="after">
+                <field name="count_attribute_line_ids" invisible="True"/>
+                <group string="Stock by companies" invisible="count_attribute_line_ids > 1">
```
dalonsod: Try this, as stock_by_company_ids has already all the information we need:
```suggestion
                <group string="Stock by companies" invisible="not stock_by_company_ids">
```

product_stock_all_companies/views/product_template_views.xml
```diff
@@ -0,0 +1,20 @@
+<odoo>
+    <record id="product_normal_form_view" model="ir.ui.view">
```
dalonsod: This ID is the same than above... copy & paste usual mistake... rename it (otherwise, only the latest view found with that name should be loaded)

dalonsod (changes requested):
See comments

dalonsod (approved):
Approved by customer :+1: 

For this version we'll keep basic internal user permission, so we're ensuring that even a pure sales user (and not inventory user) is accessing this information, and that fits better current requisites.

PR #189, opened 2025-09-04
[17.0] [MIG] stock_move_action_done_custdate_val: Migration to 17.0

lperez-solvos:
Standard migration to 17.0

PR #184, opened 2025-08-11
[17.0][MIG] stock_move_action_done_custdate: Migration to 17.0

DantePereyra:
Standard migration.

PR #190, opened 2025-09-20
[17.0][MIG] stock_valuation: Migration to 17.0

dalonsod:
TODO - VIEWS:
* [x]  SVL: move_reference vs reference
* [x] settings view
* [x] Replace readonly=True by readonly="1" in views?

TODO - DOC:
* [x] change icon
* [x] Regenerate README

Depends on:

* #184 
* #189


dalonsod:
@ChristianSantamaria this migration step should be enough, you can start other migrations that depend on it.

This migration adds a pre-migration script, not tested yet. If it arises any problem, I could temporary disable it.

dalonsod:
Enough to a first version.. let's merge

PR #191, opened 2025-09-25
[17.0][ADD] stock_secondary_unit_backorder

IriaAlonso:
When validating a delivery note and creating a partial delivery, force '_compute_secondary_uom_qty' for all lines of the partial delivery.

@dalonsod


stock_secondary_unit_move_line/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
...
+
+class StockPicking(models.Model):
+    _inherit = ["stock.picking", "product.secondary.unit.mixin"]
+    _name = "stock.picking"
+
+    def _action_done(self):
+        res = super()._action_done()
+
+        for picking in self:
+            if picking.backorder_ids:
```
dalonsod: Use `filtered` instead of `for` + `if`

stock_secondary_unit_move_line/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
...
+                for line in picking.backorder_ids.move_ids:
+                    line._compute_secondary_uom_qty()
```
dalonsod: compute method should be multi, so it could be called in that way, please try:
```suggestion
                picking.backorder_ids.move_ids._compute_secondary_uom_qty()
```

dalonsod (changes requested):
See code comments

dalonsod:
Another thing: as usual, addon name is not good... we're not updating move lines secondary unit values, but moves.

I propose `stock_secondary_unit_backorder`, what do you think?

And add to description that this is a temporary addon that is needed meanwhile base `stock_secondary_unit` addon is not fixed. Then, add a brief explanation of the current issue that is patched here with the proposed workaround.

IriaAlonso:
@dalonsod done

dalonsod (approved):
LGTM :+1: 

We'll wait for customer approval, and I'll merge it

PR #193, opened 2025-10-02
[16.0][FIX] stock_picking_deliver_to: move deliver_to field to a second line in the report

IriaAlonso:
This fix is apllied due to a general update in Odoo which the deliver_to field was not displaying correctly.

@dalonsod


dalonsod (changes requested):
Functional review :+1:

PR #200, opened 2025-12-03
[17.0][ADD] stock_picking_backorder_no_responsible

ChristianSantamaria:
Prevents the backorder picking from inheriting the responsible user from the original picking.


dalonsod:
@DantePereyra could you review?

stock_picking_backorder_no_responsible/models/stock_picking.py
```diff
@@ -0,0 +1,16 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.model_create_multi
+    def create(self, vals):
+        res = super(StockPicking, self).create(vals)
+        context = self.env.context
+        if context.get('skip_backorder') == True and context.get('cancel_backorder') == False and res.user_id:
+            res.user_id = False
+        return res
```
DantePereyra:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        context = self.env.context
        if context.get("skip_backorder") and not context.get("cancel_backorder"):
            for vals in vals_list:
                vals["user_id"] = False

        return super().create(vals_list)
```
Instead of creating the records and then updating `user_id` afterwards, it might be simpler to handle it this way.

ChristianSantamaria: The proposed `if` statement is incorrect. The `get` function returns `true` if it finds a key with the name `X` in the context. If it finds it, it will return `true`; otherwise, it will return `empty`. In other words, the return value of `get` (whether or not it finds the field) is one thing, and its value is another.

On the other hand, I think it's correct to check in the `if` statement whether the `user_id` field is filled, and if not, do nothing. Why would we clear an already empty field?

dalonsod: Regarding to `context.get()` controversy, both approaches are equivalent IMO. In fact, Dante's approach is more generic (e.g. if `skip_backorder` key value is `True`, but also `"pepe"`, it will pass `if` statement).

In the other hand, `user_id` value set... we'd prefer filling every single `vals` with proper value before calling `create`, as OCA recommends it, rather than updating already saved values.

And, at last, replace `vals` by `vals_list` at method definition, if possible.

stock_picking_backorder_no_responsible/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
...
+        for vals in vals_list:
+            picking_type_id = vals.get('picking_type_id')
+            create_backorder = False
+
+            if picking_type_id:
+                picking_type = self.env['stock.picking.type'].browse(picking_type_id)
+                create_backorder = picking_type.create_backorder == 'always'
+
+            if (
+                (context.get('skip_backorder') and not context.get('cancel_backorder'))
```
dalonsod, line 22: You can use here default value capabilities for `get` dict method (not tested):
```suggestion
                not context.get('skip_backorder', True)
```

DantePereyra:
Testing this addon with #158 and there is a combination that don't work with this addon. 
When field `create_backorder` on `picking_type` is set as `Ask` and on partner is set as `Always create`, responsible is inherited and it shouldn´t happen, could you check?

stock_picking_backorder_no_responsible/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
...
+                or create_backorder
+            ):
+                vals['user_id'] = False
+        return super(StockPicking, self).create(vals_list)
```
DantePereyra:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            if vals.get('backorder_id'):
                vals['user_id'] = False
        return super(StockPicking, self).create(vals_list)
```
Could this be the solution to problem with #158 and keep same function here?

PR #201, opened 2025-12-03
[13.0][IMP] stock_picking_mgmt_weight, logistics_planning_mgmt_weight: ticket deletion protections and permissions

ChristianSantamaria:
[IMP] stock_picking_mgmt_weight: improved deletion rules for weigh-in tickets
Modified unlink behavior to:
- Prevent deletion of weigh-in tickets with assigned classification.
- Restrict deletion of weigh-in tickets to users with proper permissions.

[IMP] logistics_planning_mgmt_weight: prevent deletion of weigh-in tickets with assigned logitics schedule

### DIFF
--- a/logistics_planning_mgmt_weight/__manifest__.py
+++ b/logistics_planning_mgmt_weight/__manifest__.py
@@ -7,7 +7,7 @@
     ''',
     'author': 'Solvos',
     'license': 'LGPL-3',
-    'version': '13.0.1.9.0',
+    'version': '13.0.1.10.0',
     'category': 'stock',
     'website': 'https://github.com/solvosci/slv-stock',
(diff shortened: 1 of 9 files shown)


dalonsod:
OK, I'll test it

dalonsod (approved):
Code & functionial review :+1: 

Only one thing that could eventually break this: `unlink` works for moves because move tickets should have a `draft` state. If a move changes (for an unexpected reason that we still don't know) to another state, unlink will fail anyway because "only draft moves can be deleted". If such case arises, a `sudo()` (or ensuring cancel + delete) will be required.

dalonsod:
I've also renamed this PR, original name was not detailed enough.

PR #203, opened 2025-12-04
[13.0][FIX] stock_production_lot_available_at_date: subtract production consumption in lot availability // cover negatives

ChristianSantamaria:
Include moves with destination usage production as consumed quantity when computing lot availability at a given date. This ensures production inputs are properly deducted from historical lot stock.

And also adds negative quant coverage (before this, only positive stocks were covered)

### DIFF
--- a/stock_production_lot_available_at_date/__manifest__.py
+++ b/stock_production_lot_available_at_date/__manifest__.py
@@ -7,7 +7,7 @@
     """,
     "author": "Solvos",
     "license": "LGPL-3",
-    "version": "13.0.1.0.0",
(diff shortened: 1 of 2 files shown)


dalonsod (approved):
Before approving this, I've added two extra features:
* Sales returns were not covered, now are also available (adding `customer` locations to `sl_src` coverage)
* By now, net negative stocks are covered, only zero stock is excluded from query.

Anyway, this feature is still incomplete, because we're not including inputs by purchases, and outputs by purchase returns. Those cases are not part of the current behavior, so we won't add them at this point.

PR #214, opened 2026-03-05
[15.0][ADD] stock_move_line_unrestricted_location

DantePereyra:
~~This module allows setting the destination location of an IN picking stock move line or the source
location of an OUT or INT picking stock move line
to any location that is a child of the warehouse of the original location.~~

_Edit_: Universal module to expand stock move location selection,
allowing all child locations of the warehouse view location.
Works for Incoming, Internal, and Outgoing pickings while keeping standard filters.


stock_move_line_unrestricted_location/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    warehouse_view_location_dest_id = fields.Many2one(
+        "stock.location",
+        related="location_dest_id.warehouse_id.view_location_id",
+        readonly=True,
+        store=False,
```
dalonsod: These two attributes actually have those values by default, when `related` is selected, so we should remove them, could you check?

stock_move_line_unrestricted_location/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+    warehouse_view_location_src_id = fields.Many2one(
+        "stock.location",
+        related="location_id.warehouse_id.view_location_id",
+        readonly=True,
+        store=False,
```
dalonsod: The same as above

stock_move_line_unrestricted_location/views/stock_move_line_views.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_stock_move_operations" model="ir.ui.view">
+        <field name="name">stock.move.operations.form (in stock_move_line_unrestricted_location_dest)</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_stock_move_operations"/>
+        <field name="priority">1001</field>
```
dalonsod: This view has no potential inheritance problems, so a lower `priority` is not needed IMO

stock_move_line_unrestricted_location/views/stock_move_line_views.xml
```diff
@@ -0,0 +1,61 @@
...
+                    <field name="warehouse_view_location_dest_id" invisible="1"/>
+                    <field name="warehouse_view_location_src_id" invisible="1"/>
+                </field>
+        </field>
+    </record>
+
+    <record id="view_stock_move_line_operation_tree" model="ir.ui.view">
+        <field name="name">stock.move.line.operations.tree (in stock_move_line_unrestricted_location_dest)</field>
+        <field name="model">stock.move.line</field>
+        <field name="inherit_id" ref="stock.view_stock_move_line_operation_tree"/>
```
dalonsod, line 8: For this view instead, a low priority is recommended, as some attributes are changed

stock_move_line_unrestricted_location/views/stock_move_line_views.xml
```diff
@@ -0,0 +1,61 @@
...
+                    ('company_id','=',company_id)]
+                </attribute>
+            </field>
+        </field>
+    </record>
+
+    <record id="view_stock_move_line_detailed_operation_tree" model="ir.ui.view">
+        <field name="name">stock.move.line.detailed.operations.tree (in stock_move_line_unrestricted_location_dest)</field>
+        <field name="model">stock.move.line</field>
+        <field name="inherit_id" ref="stock.view_stock_move_line_detailed_operation_tree"/>
```
dalonsod, line 36: The same as above

stock_move_line_unrestricted_location/views/stock_move_line_views.xml
```diff
@@ -0,0 +1,61 @@
...
+                    ('usage','!=','view'),
+                    '|', ('company_id', '=', False), ('company_id', '=', company_id)]
+                </attribute>
+            </field>
+            <field name="location_id" position="attributes">
+                <attribute name="domain">
+                    [('id','child_of',parent.warehouse_view_location_src_id),
+                    ('usage','!=','view'),
+                    ('company_id','=',company_id)]
+                </attribute>
```
dalonsod, line 27: This is right, but could you add as a XML comment the old domain that is replaced? It could help us analyzing later issues

stock_move_line_unrestricted_location/views/stock_move_line_views.xml
```diff
@@ -0,0 +1,61 @@
...
+                    ('usage','!=','view'),
+                    '|', ('company_id', '=', False), ('company_id', '=', company_id)]
+                </attribute>
+            </field>
+            <field name="location_id" position="attributes">
+                <attribute name="domain">
+                    [('id','child_of',parent.warehouse_view_location_src_id),
+                    ('usage','!=','view'),
+                    '|', ('company_id', '=', False), ('company_id', '=', company_id)]
+                </attribute>
```
dalonsod, line 55: The same as above

dalonsod (changes requested):
See initial comments, and later I'll check again.

Anyway, this is not exactly fitting the desired requirements (we need to open certain location selection... but depending on picking type). It could be complex, we could skip that requirement right now and continue

stock_move_line_unrestricted_location/views/stock_move_line_views.xml
DantePereyra: On original view, this view has `priority` 1000, so it's like this to avoid conflicts:
https://github.com/odoo/odoo/blob/15.0/addons/stock/views/stock_move_views.xml#L131

DantePereyra:
@dalonsod, the model was changed to `stock.move.line` as it was easier to access, since when the **Detailed Operations** option is enabled on the **Operation Type**, the parent view cannot be access With these changes, it is now accessible in both workflows. The rest of the requirements have been addressed:
- Changed to use `picking_type_id.warehouse_id.location_id`
- Added `priority` level on views
- Old domains left commented
- Removed `readonly=True` and `store=False` attributes as they are the default values.

dalonsod:
I still see the same code, you may not made push

dalonsod:
Only a cosmetic change to do: rename `stock_move.py` file, because current content doesn't fit with such name. Once done, I'll test it ASAP

dalonsod (approved):
Reviewed by customer 👍 

Squash made, changed commit detailed description and added myself as commit co-author.

Let's merge!

PR #215, opened 2026-03-25
[15.0][ADD] stock_account_val_operation

DantePereyra:
This addon adds new fields to the account_move_line model to link it to the stock_move, allowing to know the type of stock operation.

SLV-HT: HT01144
@dalonsod


stock_account_val_operation/models/account_move.py
```diff
@@ -0,0 +1,51 @@
...
+            elif move.origin_returned_move_id:
+                if dst_usage == 'internal':
+                    val_operation = 'in_return'
+                else:
+                    val_operation = 'out_return'
+            elif dst_usage == 'internal' and src_usage != 'internal':
+                val_operation = 'incoming'
+            elif src_usage == 'internal' and dst_usage != 'internal':
+                val_operation = 'outgoing'
+            val_operation_origin = (
```
dalonsod, line 74: Code refactoring seems to break functionality, as `val_operation` and `val_operation_origin` variables are not used to record update within for statement, could you check?

stock_account_val_operation/models/account_move.py
```diff
@@ -0,0 +1,51 @@
...
+                move.picking_id.name
+                or move.reference
+                or move.origin
+            )
+        (self - lines_with_move).write({
```
dalonsod: Inside a `compute` method never use `write`, use `update` instead:

```suggestion
        (self - lines_with_move).update({
```

For this case this won't cause any problem, but it's a good practice. E.g. for compute method of a non-stored field using `write` could lead to permission access errors.

Tip: `.` operator used at an assignment statement (like e.g. `record.field = my_value`) is an alias of `write` or `update`, depending on method context. For a `compute` method is actually an `update` statement (and for `write` for another one), so it can be used without no fear, as we've always done.

dalonsod (changes requested):
See comments. Still pending `hook` review, I'll take a look ASAP

DantePereyra:
Changes done @dalonsod . Also added a `button` on tree view to access origin when it's possible, `filter` on `search` and translations.

stock_account_val_operation/models/account_move.py
```diff
@@ -0,0 +1,69 @@
...
+                val_operation = 'incoming'
+            elif src_usage == 'internal' and dst_usage != 'internal':
+                val_operation = 'outgoing'
+            val_operation_origin = (
+                move.picking_id.name
+                or move.reference
+                or move.origin
+            )
+        line.val_operation = val_operation
+        line.val_operation_origin = val_operation_origin
```
dalonsod: Hmmm bad indentation, this won't work for multiple lines!

stock_account_val_operation/models/account_move.py
```diff
@@ -0,0 +1,69 @@
...
+            else:
+                type_operation = 'stock.picking'
+            return {
+                'type': 'ir.actions.act_window',
+                'name': _('Valuation Operation Origin'),
+                'view_mode': 'form',
+                'res_model': type_operation,
+                'target': 'new',
+                'res_id': self.env[type_operation].search([('name', '=', self.val_operation_origin)], limit=1).id,
+            }
```
dalonsod: Suggested code (prevent using `search`, as we actually have direct access to ids via ORM), there's actually a corner case that we should discuss before:

```suggestion
            res_id = False
            if self.val_operation == 'scrap':
                res_id = len(self.move_id.stock_move_id.scrap_ids) == 1 and self.move_id.stock_move_id.scrap_ids.id
            else:
                res_id = self.move_id.stock_move_id.picking_id.id
            return {
                'type': 'ir.actions.act_window',
                'name': _('Valuation Operation Origin'),
                'view_mode': 'form',
                'res_model': type_operation,
                'target': 'new',
                'res_id': res_id,
            }
```

stock_account_val_operation/models/account_move.py
```diff
@@ -0,0 +1,80 @@
...
+            line.val_operation_origin = val_operation_origin
+        (self - lines_with_move).update({
+            'val_operation': False,
+            'val_operation_origin': False,
+        })
+
+    @api.depends('val_operation')
+    def _compute_val_operation_origin_invisible(self):
+        for line in self:
+            line.val_operation_origin_invisible = not bool(line.val_operation) or line.val_operation == 'inv_adjust'
```
dalonsod: Not tested, use `filtered` instead of `for`:

```suggestion
        lines_invisible = self.filtered(lambda x: not bool(x.val_operation) or x.val_operation == 'inv_adjust')
        lines_invisible.val_operation_origin_invisible = True
        (self - lines_invisible).val_operation_origin_invisible = False
```

stock_account_val_operation/views/account_move_views.xml
```diff
@@ -0,0 +1,43 @@
...
+<odoo>
+    <record id="view_move_line_tree" model="ir.ui.view">
+        <field name="name">account.move.line.tree (in stock_account_val_operation)</field>
+        <field name="model">account.move.line</field>
+        <field name="inherit_id" ref="account.view_move_line_tree"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='move_id']" position="after">
+                <field name="val_operation" optional="show"/>
+                <field name="val_operation_origin" optional="show"/>
+                <field name="val_operation_origin_invisible" attrs="{'column_invisible': [('val_operation_origin_invisible', '=', True)]}"/>
```
dalonsod: I think this should be enough IMO (as an instrumental field, we can always hide it):

```suggestion
                <field name="val_operation_origin_invisible" column_invisible="1" />
```

dalonsod (changes requested):
See minor comments and I'll test it ASAP

stock_account_val_operation/views/account_move_views.xml
DantePereyra: This was with `attrs` (even is wrong because it should be` attrs="{'column_invisible': True}"`) but because it had this problem
```bash
2026-03-31 11:46:22,708 116262 ERROR odoo-test odoo.tools.view_validation: <string>:5:0:ERROR:RELAXNGV:RELAXNG_ERR_INVALIDATTR: Invalid attribute **column_invisible** for element field 
```

stock_account_val_operation/views/account_move_views.xml
```diff
@@ -0,0 +1,43 @@
...
+        </field>
+    </record>
+
+    <record id="view_move_line_form" model="ir.ui.view">
+        <field name="name">account.move.line.form (in stock_account_val_operation)</field>
+        <field name="model">account.move.line</field>
+        <field name="inherit_id" ref="account.view_move_line_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='name']" position="after">
+                <field name="val_operation"/>
```
dalonsod: Make invisible this when is empty

dalonsod: Still pending this

DantePereyra: Done!

dalonsod (approved):
Approved by @lromero-solvos as well, let's merge!

PR #216, opened 2026-04-07
[13.0][FIX] logistics_planning_purchase: enable initial required schedules at POL form view

dalonsod:
Before this, if a user could only edit a purchase line using inner form view, he couldn't edit initial required schedules, and eventually confirm the PO (e.g. at responsive mode). This fixes it.

HT01191


dalonsod:
:+1: at AluTest, let's merge

cc @ChristianSantamaria

